Issue #11, opened 2019-08-02
Provide capability to expose a TCP health check


tomkerkhove:
Provide the capability to expose a TCP health check.

This would allow the runtime to periodically check liveness/readiness of the host.

This should be opt-in and pulled in via a dedicated NuGet package ~ `Arcus.Messaging.Health`

tomkerkhove:
Looks like docs are only remaining, thanks @stijnmoreels!

PR #16, opened 2019-11-25
FEAT: initial health TCP probe

stijnmoreels:
Adds a TCP endpoint to probe for a health report. (initial commit)

Relates to #11


stijnmoreels:
- [ ] add YAML builds to check integration test to probe for TCP endpoint hosted in a Docker container

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+            _logger = logger;
+
+            int tcpPort = GetConfiguredTcpPort(configuration);
+            _listener = new TcpListener(IPAddress.Any, tcpPort);
+            _serializerSettings = CreateDefaultSerializerSettings();
+        }
+
+        private static int GetConfiguredTcpPort(IConfiguration configuration)
+        {
+            const string arcusTcpPort = "ARCUS_TCP_PORT";
```
tomkerkhove: I'd go with `HEALTH` to explicitly show that it will only be used for that
```suggestion
            const string arcusTcpPort = "ARCUS_HEALTH_PORT";
```

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Arcus.Messaging.Health.Tcp
+{
+    /// <summary>
+    /// Representing a TCP server as a background process to expose an endpoint where the <see cref="HealthReport"/> is being broadcasted.
+    /// </summary>
+    public class TcpHealthServer : BackgroundService
```
tomkerkhove: Suggestion, not required

```suggestion
    public class TcpHealthListener : BackgroundService
```

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+            string tcpPortString = configuration[arcusTcpPort];
+            var tcpPort = 0;
+            Guard.For<ArgumentException>(
+                () => !Int32.TryParse(tcpPortString, out tcpPort), 
+                $"Requires a configuration implementation with a '{arcusTcpPort}' key containing a TCP port number");
+
+            return tcpPort;
+        }
+
+        private static JsonSerializerSettings CreateDefaultSerializerSettings()
```
tomkerkhove: We might want to make this configurable/overridable? 🤔

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+        public override Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return Task.FromCanceled(cancellationToken);
+            }
+
+            try
+            {
+                _logger.LogTrace("Starting TCP server...");
```
tomkerkhove: Let's include the port and use structured logging

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+                _listener.Start();
+                _logger.LogInformation("TCP server started!");
```
tomkerkhove: Let's include the port as well and use structured logging

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+        public TcpHealthServer(IConfiguration configuration, HealthCheckService healthService, ILogger<TcpHealthServer> logger)
+        {
+            Guard.NotNull(configuration, nameof(configuration), "Requires a configuration implementation to retrieve hosting information for the TCP server");
+            Guard.NotNull(healthService, nameof(healthService), "Requires a health service to retrieve the current health status of the application");
+            Guard.NotNull(logger, nameof(logger), "Requires a logger implementation to write diagnostic messages during the running of the TCP server");
+
+            _healthService = healthService;
+            _logger = logger;
+
+            int tcpPort = GetConfiguredTcpPort(configuration);
```
tomkerkhove: Let's maybe capture this in a `Port` property?

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// This method is called when the <see cref="T:Microsoft.Extensions.Hosting.IHostedService" /> starts. The implementation should return a task that represents
+        /// the lifetime of the long running operation(s) being performed.
+        /// </summary>
+        /// <param name="stoppingToken">Triggered when <see cref="M:Microsoft.Extensions.Hosting.IHostedService.StopAsync(System.Threading.CancellationToken)" /> is called.</param>
+        /// <returns>A <see cref="T:System.Threading.Tasks.Task" /> that represents the long running operations.</returns>
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        {
+            while (!stoppingToken.IsCancellationRequested)
+            {
+                _logger.LogInformation("Worker running at: {time}", DateTimeOffset.Now);
```
tomkerkhove: Let's remove this maybe as it doesn't have much value?

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+
+                await AcceptConnectionAsync();
+            }
+        }
+
+        private async Task AcceptConnectionAsync()
+        {
+            _logger.LogTrace("Accepting TCP client...");
```
tomkerkhove: Let's add the port here as well

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+            using (TcpClient client = await _listener.AcceptTcpClientAsync())
+            {
+                _logger.LogInformation("TCP client accepted!");
+                using (NetworkStream ns = client.GetStream())
```
tomkerkhove:
```suggestion
                using (NetworkStream networkStream = client.GetStream())
```

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+                {
+                    // TODO: something to serialize as BSON instead?
```
tomkerkhove: As discussed, I would not do this for now

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+                    HealthReport report = await _healthService.CheckHealthAsync();
+                    string serialized = JsonConvert.SerializeObject(report, _serializerSettings);
```
tomkerkhove: Thought, would it make sense to log the health report, or just it's status? 🤔

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+
+                    byte[] response = Encoding.UTF8.GetBytes(serialized);
+                    ns.Write(response, 0, response.Length);
+                    _logger.LogInformation("Return health report: {status}", report.Status);
```
tomkerkhove: Maybe we should add some info concerning the client?

tomkerkhove: Let's use capitals in case folks use things like Application Insights

```suggestion
                    _logger.LogInformation("Return health report: {Status}", report.Status);
```

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
```diff
@@ -0,0 +1,161 @@
...
+        public override Task StopAsync(CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return Task.FromCanceled(cancellationToken);
+            }
+
+            try
+            {
+                _logger.LogTrace("Stopping TCP server...");
```
tomkerkhove: Again, port

tomkerkhove: Main reason is that you don't always have that context here if you browse the logs

src/Arcus.Messaging.Tests/Health/TcpHealthCheckTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Arcus.Messaging.Tests.Health
+{
+    public class TcpHealthCheckTests
+    {
+        [Fact]
+        public async Task TcpHealthServer_ProbeForHealthReport_ResponseHealthy()
```
tomkerkhove: Did this work in the end?

src/docker-compose.override.yml
```diff
@@ -0,0 +1,8 @@
+version: '3.4'
+
+services:
+  arcus.messaging.tests.worker:
+    environment:
+      - ARCUS_TCP_PORT=5001
```
tomkerkhove: Align with suggested rename here maybe

tomkerkhove:
> * [ ]  add YAML builds to run integration test to probe for TCP endpoint hosted in a Docker container

Let's already have a look at building the same job against .NET Core 2.2 & 3.0 maybe?
This might probably be what we need https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=example#template-references

But it's ok to do this as a follow-up thought

src/Arcus.Messaging.Tests/Health/TcpHealthCheckTests.cs
stijnmoreels: Yep! 🎊

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: Yeah, or that we can override to something not-JSON. But anyway, furture PR's :)

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: The F# dev in me :)

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: The status is already logged. The entire report (w/ entries and everything) is maybe too much.

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: Like what for example? Port?

stijnmoreels:
> > * [ ]   add YAML builds to run integration test to probe for TCP endpoint hosted in a Docker container
> 
> Let's already have a look at building the same job against .NET Core 2.2 & 3.0 maybe?
> This might probably be what we need https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=example#template-references
> 
> But it's ok to do this as a follow-up thought

I've created a YAML build that builds the project now in a .NET Core 3.0 env (inc. Docker).

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: Hmm, listener/speaker, client/server, sender/receiver, ... not sure.

src/Arcus.Messaging.Tcp.Orchestrator.dcproj
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
stijnmoreels: Done

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
tomkerkhove: Server sounds too robust imo as it just listens for connections and not use TCP for the whole WebAPI, but that's just my opinion

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
tomkerkhove: If you're parking it, then I would suggest to create an issue for it

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
tomkerkhove: It's already logged, but only when it was serialized & sent back. Maybe we want to move that up here before we serialize it (I'm very skeptical about waiting that long to log it, but maybe just me)

src/Arcus.Messaging.Tcp.Orchestrator.dcproj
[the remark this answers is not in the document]
tomkerkhove: Are you sure? It still says `Arcus.Messaging.Tcp.Orchestrator` 🤔

tomkerkhove:
Thanks, I presume you'll configure the build for this repo once it's merged?

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
tomkerkhove: Well it doesn't say a lot, sorry :)

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
tomkerkhove: Something that could help troubleshooting like an identifier or so

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: Ah, you're right. It's better to move. Thanks.

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: #18 added

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: Renamed.

src/Arcus.Messaging.Health/Tcp/TcpHealthServer.cs
stijnmoreels: Used `client.Socket.Endpoint`.

stijnmoreels:
I think this is done, can you check @tomkerkhove ? I'll update the source of the build in devops afterwards.

PR #19, opened 2019-12-04
Provide a convenient way to opt-in for TCP health probing

tomkerkhove:
Provide a convenient way to opt-in for TCP health probing.

- Introduce `AddTcpHealthProbes` to opt-in and allow you to register health checks, if you want

```csharp
.ConfigureServices((hostContext, services) =>
{
    services.AddTcpHealthProbes(healthCheckBuilder =>
    {
        healthCheckBuilder.AddCheck("sample", () => HealthCheckResult.Healthy());
    });
});
```

```csharp
.ConfigureServices((hostContext, services) =>
{
    services.AddTcpHealthProbes();
    services.AddHealthChecks()
        .AddCheck("sample", () => HealthCheckResult.Healthy());
});
```

Relates to #11

### DIFF
--- a/src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
+++ b/src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
@@ -23,13 +23,10 @@ namespace Microsoft.Extensions.DependencyInjection
         {
             Guard.NotNull(services, nameof(services));
 
-            services.AddHostedService<TcpHealthListener>();
+            var healthCheckBuilder = services.AddHealthChecks();
+            configureHealthChecks?.Invoke(healthCheckBuilder);
 
-            if (configureHealthChecks != null)
-            {
-                var healthCheckBuilder = services.AddHealthChecks();
-                configureHealthChecks(healthCheckBuilder);
-            }
+            services.AddHostedService<TcpHealthListener>();
 
             return services;
         }
--- a/src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
+++ b/src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
@@ -1,0 +1,37 @@
+﻿using System;
+using Arcus.Messaging.Health.Tcp;
+using GuardNet;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.DependencyInjection
+{
+    // ReSharper disable once InconsistentNaming
+    public static class IServiceCollectionExtensions
+    {
+        /// <summary>
+        ///     Adds TCP health probing
+        /// </summary>
+        /// <remarks>
+        ///     In order for the TCP health probes to work, ASP.NET Core Health Checks must be configured. You can configure
+        ///     it here or do it yourself as well.
+        /// </remarks>
+        /// <param name="services">Collection of services to use in the application</param>
+        /// <param name="configureHealthChecks">Capability to configure the required health checks to expose, if required</param>
+        /// <returns>Collection of services to use in the application</returns>
+        public static IServiceCollection AddTcpHealthProbes(this IServiceCollection services,
+            Action<IHealthChecksBuilder> configureHealthChecks = null)
+        {
+            Guard.NotNull(services, nameof(services));
+
+            services.AddHostedService<TcpHealthListener>();
+
+            if (configureHealthChecks != null)
+            {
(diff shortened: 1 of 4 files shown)


src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        public static IServiceCollection AddTcpHealthProbes(this IServiceCollection services,
+            Action<IHealthChecksBuilder> configureHealthChecks = null)
+        {
+            Guard.NotNull(services, nameof(services));
+
+            services.AddHostedService<TcpHealthListener>();
+
+            if (configureHealthChecks != null)
+            {
+                var healthCheckBuilder = services.AddHealthChecks();
```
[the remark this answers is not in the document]
tomkerkhove: That would also simplify this:
```csharp
.ConfigureServices((hostContext, services) =>
{
    services.AddTcpHealthProbes();
    services.AddHealthChecks()
        .AddCheck("sample", () => HealthCheckResult.Healthy());
});
```

To:
```csharp
.ConfigureServices((hostContext, services) =>
{
    services.AddTcpHealthProbes();
});
```

tomkerkhove:
Fixing test as we speak, but is strange one 😅

image

src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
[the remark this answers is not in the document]
stijnmoreels: Makes sense, altough ASP.NET Core/.NET Core extensions are full of these dependencies that has to be registered in the correct sequence. That being said, I think it certainly would be an added value to have a single configuration line for this.

src/Arcus.Messaging.Tests.Integration/Health/TcpHealthListenerTests.cs
```diff
@@ -34,7 +35,7 @@ public async Task TcpHealthServer_ProbeForHealthReport_ResponseHealthy()
                 using (var reader = new StreamReader(clientStream))
                 {
                     // Act
-                    string healthReport = await reader.ReadLineAsync();
+                    string healthReport = await reader.ReadToEndAsync();
```
stijnmoreels: You had some problems with this?

src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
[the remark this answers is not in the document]
stijnmoreels: Also, we should first register the health services before our TCP probe, no?

src/Arcus.Messaging.Tests.Integration/Health/TcpHealthListenerTests.cs
tomkerkhove: I was figuring out #21 so can be removed, but think it's always best to make sure we've read to the end, no? What if there's a new line in there for example, won't that break it?

src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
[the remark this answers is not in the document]
tomkerkhove: > Also, we should first register the health services before our TCP probe, no?

Good point!

src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -23,13 +23,10 @@ public static class IServiceCollectionExtensions
         {
             Guard.NotNull(services, nameof(services));
 
-            services.AddHostedService<TcpHealthListener>();
+            var healthCheckBuilder = services.AddHealthChecks();
```
stijnmoreels: 👍

src/Arcus.Messaging.Tests.Integration/Health/TcpHealthListenerTests.cs
stijnmoreels: Yeah, right. Was just thinking that with an infinite/gigantic input; it could cause availability issues when there isn't any maximum of some sort.
Just thinking out loud; but this is a test so, less related I guess.

tomkerkhove: It should be fine indeed I think!

PR #40, opened 2019-12-18
DOC: add feature docs for TCP health probe

stijnmoreels:
Adds feature docs for the TCP health probe.

Relates to #11

Ps.: the folder structure will probably change when more features are added; so I added this feature docs at the root of the `./features` folder.

### DIFF
--- a/docs/features/tcp-health-probe.md
+++ b/docs/features/tcp-health-probe.md
@@ -5,7 +5,7 @@ layout: default
 
 # TCP Health probe
 
-We provide a TCP health probe endpoint that allows a runtime to periodically check the liveness/readiness of the host based on the .NET Core health approach.
+We provide a TCP health probe endpoint that allows a runtime to periodically check the liveness/readiness of the host based on the [.NET Core health checks](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks).
 
 ## Installation
 
--- a/docs/features/tcp-health-probe.md
+++ b/docs/features/tcp-health-probe.md
@@ -5,7 +5,7 @@ layout: default
 
 # TCP Health probe
 
-A `BackgroundService` can be added to the <span>ASP.NET</span> hosted services to expose a TCP health endpoint that allows a runtime to periodically check the liveness/readiness of the host.
+We provide a TCP health probe endpoint that allows a runtime to periodically check the liveness/readiness of the host based on the .NET Core health approach.
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -12,7 +12,7 @@ Coming soon!
 
 # Features
 
-Coming soon!
+- [TCP health probe](features/tcp-health-probe) allow the runtime to periodically check liveness/readiness of the host.
```
tomkerkhove: How about this?
```suggestion
- Support for exposing TCP health probes to periodically check liveness/readiness of the host ([docs](features/tcp-health-probe))
```

docs/features/tcp-health-probe.md
```diff
@@ -0,0 +1,40 @@
...
+## Usage
+
+To include the TCP endpoint, add the following line of code in the `Startup.ConfigureServices` method:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddTcpHealthProbes(healthBuilder => 
+    {
+        healthBuilder.AddCheck("Example", () => HealthCheckResult.Healthy("Example is OK!"), tags: new[] { "example" })
```
tomkerkhove: Maybe mention that this is optional and you can just do `services.AddTcpHealthProbes()`?

docs/features/tcp-health-probe.md
```diff
@@ -0,0 +1,40 @@
+---
+title: "TCP Health probe"
+layout: default
+---
+
+# TCP Health probe
+
+A `BackgroundService` can be added to the <span>ASP.NET</span> hosted services to expose a TCP health endpoint that allows a runtime to periodically check the liveness/readiness of the host.
```
tomkerkhove: I'm not sure if we have to go so technical, isn't it just enough to say that we provide a TCP health probe endpoint which will expose health information based on .NET Core health approach?

_PS: It's not ASP.NET here so would leave that out maybe?_

docs/index.md
stijnmoreels: Ah, yes; was thinking about the approach used in the other repo's.

docs/features/tcp-health-probe.md
stijnmoreels: Ah, right; I'll change this.

docs/features/tcp-health-probe.md
stijnmoreels: Yep, added this in the inline code now, with comments explaining.

docs/features/tcp-health-probe.md
```diff
@@ -0,0 +1,44 @@
+---
+title: "TCP Health probe"
+layout: default
+---
+
+# TCP Health probe
+
+We provide a TCP health probe endpoint that allows a runtime to periodically check the liveness/readiness of the host based on the .NET Core health approach.
```
tomkerkhove:
```suggestion
We provide a TCP health probe endpoint that allows a runtime to periodically check the liveness/readiness of the host based on the [.NET Core health checks](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks).
```

stijnmoreels: Ah, yes. Thanks!

Issue #21, opened 2019-12-04
Health report serialization doesn't work with individual checks


tomkerkhove:
Health report serialization doesn't work with individual checks as they are not correctly deserialized.

As part of our `TcpHealthListener.AcceptConnectionAsync` we serialize the `HealthReport` and send it back to the TCP client for interpretation.

It turns out that deserializing that payload, even in `TcpHealthListener.AcceptConnectionAsync`, doesn't work correctly if you have configured any `AddCheck` dependencies. Reason is that `HealthReportEntry` is not being deserialized correctly which causes the whole report to be marked as `Unhealthy` (enum default).

**Important:** The serialization settings should be as minimal as possible so that consumers can just use it.

## Repro

1. Configure TCP listener & probes, here via our extension method:

```csharp
services.AddTcpHealthProbes(healthCheckBuilder =>
{
    healthCheckBuilder.AddCheck("sample", () => HealthCheckResult.Healthy());
});
```

2. Ping TCP port for report or get it yourself

3. Serialize
```csharp
string serialized = JsonConvert.SerializeObject(report, _serializerSettings);
```

Result will be similar to:
```json
{
	"Entries": {
		"sample": {
			"Data": {},
			"Duration": "00:00:00.0017537",
			"Status": "Healthy",
			"Tags": []
		}
	},
	"Status": "Healthy",
	"TotalDuration": "00:00:00.0217830"
}
```

4. Deserialize, even with same serialization settings it fails
```csharp
var report = JsonConvert.DeserializeObject<HealthReport>(serialized, _serializerSettings);
```

Result will not be correct:
![image](https://user-images.githubusercontent.com/4345663/70126927-4b196c80-167a-11ea-9e6d-544b0f849f1e.png)

tomkerkhove:
I've changed it from bug to enhancement because it's not a bug in our platform but merely how we have to consume it as a client (in the tests/real-world)

PR #41, opened 2019-12-31
FIX: health report deserialization with individual check

stijnmoreels:
Fixes how we deserializat the `HealthReport`.
The problem exists in that the report is made for serialization but not _deserialization_. The entries are `struct`'s and don't get deserialized correctly.

This could be fixed with a dedicated `JsonConverter` that get passed with the deserialization call.

Relates to #21

### DIFF
--- a/src/Arcus.Messaging.Tests.Integration/Health/HealthReportEntryConverter.cs
+++ b/src/Arcus.Messaging.Tests.Integration/Health/HealthReportEntryConverter.cs
@@ -9,6 +9,8 @@ namespace Arcus.Messaging.Tests.Integration.Health
 {
     /// <summary>
     /// JSON converter to deserialize <see cref="HealthReportEntry"/> <c>struct</c>.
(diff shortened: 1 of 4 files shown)


stijnmoreels:
The tests in the repo where the `HealthReport` lives; also uses directly the `string` representation and not the original model for this.
Makes sense in a way; that the health report model is not exposed.

src/Arcus.Messaging.Tests.Integration/Health/HealthReportEntryConverter.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Arcus.Messaging.Tests.Integration.Health
+{
+    /// <summary>
+    /// JSON converter to deserialize <see cref="HealthReportEntry"/> <c>struct</c>.
```
tomkerkhove: Maybe add a small comment here why we are doing this for others?

PR #42, opened 2020-01-03
FEAT: make HealthReport serialization & TCP health port configurable

### LINKED ISSUE #18: Make serialization of health report configurable
### Current Situation
The current health report always gets serialized to a JSON representation using fixed settings.

### Proposed Situation
Make the serialization logic overridable so users can choose there own format and settings.

### LINKED ISSUE #20: Make TCP port of health report configurable via options
Make TCP port of health report configurable via options a la `TcpHealthListenerOptions` in case they do not want to use configuration.

stijnmoreels:
Adds a `TcpHealthListenerOptions` to the services collection so we can control the serialization of the `HealthReport` and the TCP port of the health check endpoint.

Closes #18 
Closes #20

### DIFF
--- a/docs/features/tcp-health-probe.md
+++ b/docs/features/tcp-health-probe.md
@@ -23,13 +23,15 @@ To include the TCP endpoint, add the following line of code in the `Startup.Conf
 public void ConfigureServices(IServiceCollection services)
 {
     // Add TCP health probe without extra health checks.
-    services.AddTcpHealthProbes();
+    services.AddTcpHealthProbes("MyConfigurationKeyToTcpHealthPort");
 
     // Or, add your extra health checks in a configuration delegate.
-    services.AddTcpHealthProbes(healthBuilder => 
-    {
-        healthBuilder.AddCheck("Example", () => HealthCheckResult.Healthy("Example is OK!"), tags: new[] { "example" })
-    });
+    services.AddTcpHealthProbes(
+        "MyConfigurationkeyToTcpHealthPort",
+        configureHealthChecks: healthBuilder => 
+        {
+            healthBuilder.AddCheck("Example", () => HealthCheckResult.Healthy("Example is OK!"), tags: new[] { "example" })
+        });
 }
 ```
 
@@ -42,6 +44,7 @@ public void ConfigureServices(IServiceCollection services)
 {
     // Add TCP health probe with or whitout extra health checks.
     services.AddTcpHealthProbes(
+        "MyConfigurationKeyToTcpHealthPort",
         configureTcpListenerOptions: options =>
         {
             // Configure the configuration key on which the health report is exposed.
--- a/docs/features/tcp-health-probe.md
+++ b/docs/features/tcp-health-probe.md
@@ -35,10 +35,30 @@ public void ConfigureServices(IServiceCollection services)
 
 ## Configuration
 
-To make the TCP health check fully functional, you'll  require some configuration values. 
+The TCP probe allows several additional configuration options.
 
-| Configuration key   | Usage        | Type  | Description                                                         |
-| ------------------- | ------------ | ----- | ------------------------------------------------------------------- |
-| `ARCUS_HEALTH_PORT` | **required** | `int` | The TCP port on which the health report is exposed.                 |
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    // Add TCP health probe with or whitout extra health checks.
+    services.AddTcpHealthProbes(
+        configureTcpListenerOptions: options =>
+        {
+            // Configure the configuration key on which the health report is exposed.
+            options.TcpPortConfigurationKey = "MyConfigurationKey";
+
+            // Configure how the health report should be serialized.
+            options.HealthReportSerializer = new MyHealthReportSerializer();
+        });
+}
+
+public class MyHealthReportSerializer : IHealthReportSerializer
+{
+    public byte[] Serialize(HealthReport healthReport)
+    {
+        return Array.Empty<byte>();
+	}
+}
+```
 
 [&larr; back](/)
--- a/src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
+++ b/src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
@@ -17,14 +17,22 @@ namespace Microsoft.Extensions.DependencyInjection
         /// </remarks>
         /// <param name="services">Collection of services to use in the application</param>
         /// <param name="configureHealthChecks">Capability to configure the required health checks to expose, if required</param>
+        /// <param name="configureTcpListenerOptions">Capability to configure additional options how the <see cref="TcpHealthListener"/> works, if required</param>
         /// <returns>Collection of services to use in the application</returns>
-        public static IServiceCollection AddTcpHealthProbes(this IServiceCollection services,
-            Action<IHealthChecksBuilder> configureHealthChecks = null)
+        public static IServiceCollection AddTcpHealthProbes(
+            this IServiceCollection services,
+            Action<IHealthChecksBuilder> configureHealthChecks = null,
+            Action<TcpHealthListenerOptions> configureTcpListenerOptions = null)
         {
             Guard.NotNull(services, nameof(services));
 
             var healthCheckBuilder = services.AddHealthChecks();
             configureHealthChecks?.Invoke(healthCheckBuilder);
+
+            if (configureTcpListenerOptions != null)
+            {
+                services.Configure(configureTcpListenerOptions);
+            }
 
             services.AddHostedService<TcpHealthListener>();
 
--- a/src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
+++ b/src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
@@ -16,23 +16,27 @@ namespace Microsoft.Extensions.DependencyInjection
         ///     it here or do it yourself as well.
         /// </remarks>
         /// <param name="services">Collection of services to use in the application</param>
+        /// <param name="tcpConfigurationKey">The TCP health port on which the health report is exposed.</param>
         /// <param name="configureHealthChecks">Capability to configure the required health checks to expose, if required</param>
         /// <param name="configureTcpListenerOptions">Capability to configure additional options how the <see cref="TcpHealthListener"/> works, if required</param>
         /// <returns>Collection of services to use in the application</returns>
         public static IServiceCollection AddTcpHealthProbes(
             this IServiceCollection services,
+            string tcpConfigurationKey,
             Action<IHealthChecksBuilder> configureHealthChecks = null,
             Action<TcpHealthListenerOptions> configureTcpListenerOptions = null)
         {
-            Guard.NotNull(services, nameof(services));
+            Guard.NotNull(services, nameof(services), "Requires a set of services to add the TCP health probe to");
+            Guard.NotNullOrWhitespace(tcpConfigurationKey, nameof(tcpConfigurationKey), "Requires a non-blank configuration key to retrieve the TCP port");
 
             var healthCheckBuilder = services.AddHealthChecks();
             configureHealthChecks?.Invoke(healthCheckBuilder);
 
-            if (configureTcpListenerOptions != null)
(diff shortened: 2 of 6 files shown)


src/Arcus.Messaging.Health/Tcp/TcpHealthListener.cs
```diff
@@ -127,8 +119,7 @@ private async Task AcceptConnectionAsync()
                     string clientId = client.Client?.RemoteEndPoint?.ToString() ?? String.Empty;
                     _logger.LogInformation("Return '{Status}' health report to client {ClientId}", report.Status, clientId);
 
-                    string serialized = JsonConvert.SerializeObject(report, _serializerSettings);
-                    byte[] response = Encoding.UTF8.GetBytes(serialized);
+                    byte[] response = _tcpListenerOptions.ReportSerializer(report);
```
tomkerkhove: Do we use this approach, or should people assign a different serializer to our options?

Ie, `TcpHealthListenerOptions.Serializer` could be our default unless they change it to another one

stijnmoreels: The default `ReportSerializer` is set as default, and will be used if the user doesn't change it with `.Configure(options => options.ReportSerializer = ...)` .

tomkerkhove: I know, but now it's a function - Doesn't it feel more natural if you just give it a deserializer rather than implementation?

stijnmoreels: :D FP vs OOP.
`ISerializer` is a .NET interface I think? Would you rather have that we use that one?

tomkerkhove: It might be just me, what do you think @fgheysels?

stijnmoreels: Hmm, the message pump has also some functions instead of objects for settings, we should consider use the same approach?

tomkerkhove: That's different imo - The message pump allows you to configure the message pump by mutating the default/existing configuration by just changing the object.

Here we are doing a whole lot of processing by passing a lambda, this is not really OOP as you've mentioned

tomkerkhove:
Would you mind updating the docs on this when the implementation is agreed upon please?

stijnmoreels:
> Would you mind updating the docs on this when the implementation is agreed upon please?

Ah, yes. Lets keep this comment open until that's done.

src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -17,15 +17,23 @@ public static class IServiceCollectionExtensions
         /// </remarks>
         /// <param name="services">Collection of services to use in the application</param>
         /// <param name="configureHealthChecks">Capability to configure the required health checks to expose, if required</param>
+        /// <param name="configureTcpListenerOptions">Capability to configure additional options how the <see cref="TcpHealthListener"/> works, if required</param>
         /// <returns>Collection of services to use in the application</returns>
-        public static IServiceCollection AddTcpHealthProbes(this IServiceCollection services,
-            Action<IHealthChecksBuilder> configureHealthChecks = null)
+        public static IServiceCollection AddTcpHealthProbes(
```
stijnmoreels: Ah, yes. Lets keep this comment open until that's done.

src/Arcus.Messaging.Health/Tcp/TcpHealthListener.cs
stijnmoreels: So introducing two interfaces and two default implementations?

tomkerkhove: Why two?

stijnmoreels: To retrieve the TCP health port and to set the serialization logic to parse health reports.

tomkerkhove: TCP health port should just be a configuration key name in the extension such as:
```csharp
services.AddTcpHealthProbes("ARCUS_HEALTH_PORT");
```

This should be more than enough for now, if people need more control we can still open it up later on!

Commented this suggestion somewhere this morning but can't seem to find it anymore :/

src/Arcus.Messaging.Health/Tcp/TcpHealthListenerOptions.cs
```diff
@@ -68,15 +54,15 @@ private static JsonSerializerSettings CreateDefaultSerializerSettings()
         }
 
         /// <summary>
-        /// Gets or sets the function that gets the TCP health check port, using the service object with a registered collection of instances.
+        /// Gets or sets the TCP health port on which the health report is exposed.
         /// </summary>
-        public Func<IServiceProvider, int> GetTcpHealthPort
+        public string TcpHealthPort
```
tomkerkhove: Since it's called `TcpHealthListenerOptions`, I'd simply call this `PortConfigurationKey` or so?

src/Arcus.Messaging.Health/Tcp/TcpHealthListener.cs
tomkerkhove: For the other one an interface would be sufficient for me. If the `CustomSerializer` or so is not specified we can use the default implementation without the additional wrapper.

Don't think we should create an default implementation which just calls our dependent library imo but I'm open to anything

stijnmoreels: Added a suggestion.

src/Arcus.Messaging.Health/Tcp/TcpHealthListenerOptions.cs
stijnmoreels: Renamed to `TcpPortConfigurationKey`

docs/features/tcp-health-probe.md
```diff
@@ -35,10 +35,30 @@ public void ConfigureServices(IServiceCollection services)
...
+The TCP probe allows several additional configuration options.
 
-| Configuration key   | Usage        | Type  | Description                                                         |
-| ------------------- | ------------ | ----- | ------------------------------------------------------------------- |
-| `ARCUS_HEALTH_PORT` | **required** | `int` | The TCP port on which the health report is exposed.                 |
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    // Add TCP health probe with or whitout extra health checks.
+    services.AddTcpHealthProbes(
```
tomkerkhove: Do we use a default `TcpPortConfigurationKey` or make it a mandatory parameter here? I'd say the latter to be honest.

src/Arcus.Messaging.Health/Tcp/TcpHealthListenerOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// Options to control how the <see cref="TcpHealthListener"/> acts as a TCP endpoint for the <see cref="HealthReport"/>.
+    /// </summary>
+    public class TcpHealthListenerOptions
+    {
+        private string _tcpHealthPort = "ARCUS_HEALTH_PORT";
+
+        /// <summary>
+        /// Gets or sets the function that serializes the <see cref="HealthReport"/> to a series of bytes.
+        /// </summary>
+        public IHealthReportSerializer HealthReportSerializer { get; set; }
```
tomkerkhove: Maybe just call it `Serializer`? We're already in the TCP health context.

src/Arcus.Messaging.Health/Tcp/TcpHealthListener.cs
```diff
@@ -137,6 +144,29 @@ private async Task AcceptConnectionAsync()
             }
         }
 
+        private byte[] SerializeHealthReport(HealthReport healthReport)
+        {
+            IHealthReportSerializer reportSerializer = _tcpListenerOptions.HealthReportSerializer;
+            if (reportSerializer is null)
```
tomkerkhove: THinking out loud - If with 2 implementations, split it in two methods or not?

docs/features/tcp-health-probe.md
```diff
@@ -35,10 +35,30 @@ public void ConfigureServices(IServiceCollection services)
...
+        {
+            // Configure the configuration key on which the health report is exposed.
+            options.TcpPortConfigurationKey = "MyConfigurationKey";
+
+            // Configure how the health report should be serialized.
+            options.HealthReportSerializer = new MyHealthReportSerializer();
+        });
+}
+
+public class MyHealthReportSerializer : IHealthReportSerializer
```
tomkerkhove: I would move this to a second paragraph so that we have:
1. A section on minimal required
2. A section on bringing your own serialization

stijnmoreels: ? But isn't that already the case? WIth the `Usage` section?

src/Arcus.Messaging.Health/Tcp/TcpHealthListener.cs
stijnmoreels: Normally, yes. But I think in this case the implementation is too small to really benefit from it?

src/Arcus.Messaging.Health/Tcp/TcpHealthListenerOptions.cs
stijnmoreels: Yeah, but in the lamdba it would look like `options => options` I think it adds more description this way?

docs/features/tcp-health-probe.md
stijnmoreels: Yep, done.

src/Arcus.Messaging.Health/Tcp/TcpHealthListener.cs
tomkerkhove: 👌

docs/features/tcp-health-probe.md
tomkerkhove: I've missed that one, sorry!

src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -16,16 +16,28 @@ public static class IServiceCollectionExtensions
         ///     it here or do it yourself as well.
         /// </remarks>
         /// <param name="services">Collection of services to use in the application</param>
+        /// <param name="tcpConfigurationKey">The TCP health port on which the health report is exposed.</param>
```
tomkerkhove: Nitpick - Maybe be a bit more explicit?
```suggestion
        /// <param name="tcpConfigurationKey">Configuration key that defined the TCP health port on which the health report is exposed.</param>
```

src/Arcus.Messaging.Health/Tcp/TcpHealthListenerOptions.cs
tomkerkhove: I presume you are talking about the registration then?

```csharp
services.AddTcpHealthProbes("MyConfigurationKey",
        configureTcpListenerOptions: options =>
        {
            // Configure how the health report should be serialized.
            options.HealthReportSerializer = new MyHealthReportSerializer();
        });
```

```csharp
services.AddTcpHealthProbes("MyConfigurationKey",
        configureTcpListenerOptions: options =>
        {
            // Configure how the health report should be serialized.
            options.Serializer = new MyHealthReportSerializer();
        });
```

I think both are fine, just *personally* think `HealthReport` doesn't have much added value but up to you!

src/Arcus.Messaging.Health/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Yep, done.

src/Arcus.Messaging.Health/Tcp/TcpHealthListenerOptions.cs
stijnmoreels: I've changed it.

tomkerkhove: Thanks!

tomkerkhove:
LGTM, thanks @stijnmoreels !

Issue #35, opened 2019-12-16
Provide seperation between concepts of `message pump` & `message handler`


tomkerkhove:
Provide seperation between concepts of `message pump` & `message handler` instead of mixing both in the current state.

The message pump should get all messages and hand them over to the message handler to process them.

stijnmoreels:
I like the idea. So I guess the `MessagePump.ProcessMessageAsync` will become an interface?

```csharp
protected abstract Task ProcessMessageAsync(TMessage message, TMessageContext messageContext,
            MessageCorrelationInfo correlationInfo, CancellationToken cancellationToken);
```

We could start with this little change: move it to a single implementation, then we could later see how #27 would work out?

stijnmoreels:
Something like this, in the `MessagePump.cs`

```csharp
protected async Task ProcessMessageAsync(
    TMessage message,
    TMessageContext messageContext,
    MessageCorrelationInfo correlationInfo,
    CancellationToken cancellationToken)
{
    var messageHandler = ServiceProvider.GetRequiredService<IMessageHandler<TMessage, TMessageContext>>();
    await messageHandler.ProcessMessageAsync(message, messageContext, correlationInfo, cancellationToken);
}
```

tomkerkhove:
Interesting approach and good idea, would you mind creating a PR with this as a POC?

stijnmoreels:
> Interesting approach and good idea, would you mind creating a PR with this as a POC?

Pushed some stuff. And also made an addition to how diff message handlers could be selected based on the messaging context.

Just all ideas, nothing serious.

stijnmoreels:
Keeping open till docs are done...

PR #49, opened 2020-01-27
Feature - provide seperation between concepts of `message pump` & `message handler`

stijnmoreels:
Relates to #35


src/Arcus.Messaging.Tests.Workers.ServiceBus.Queue/Program.cs
```diff
@@ -36,7 +37,9 @@ public static void Main(string[] args)
                             .UsingAuthenticationKey(eventGridKey)
                             .Build();
                     });
-                    services.AddServiceBusQueueMessagePump<OrdersMessagePump>(configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING"]);
+                    services.AddServiceBusQueueMessagePump<OrdersMessagePump>(configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING"])
```
tomkerkhove: I presume we don't need the message pump type here anymore?
```suggestion
                    services.AddServiceBusQueueMessagePump(configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING"])
```

stijnmoreels: No, we _do_ need that for the registration of the message pump.
The other calls registers the handlers.

tomkerkhove: We need the Service Bus message pump, but that's it, the message handler is implementation specific.

tomkerkhove: That's the goal of #35

stijnmoreels: Yes, I think I get the goal, only not sure how this relates to this.
There's nothing changed on the way the message pump is added/registered, only the handlers are now processing the messages instead of the message pump.
So, we need to register both the pump and the handlers.

tomkerkhove: What I expected to see would be:
```csharp
services.AddServiceBusQueueMessagePump(configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING"])
        .WithMessageHandler<Order, OrdersMessageHandler>();
```

What would the `messageContext => true` be for? Filtering? If so, I'd leave that out for now but that's my personal opinion.

Something we should provide is a `WithFallbackMessageHandler` in case the above flow doesn't find a matching handler, but that might be more interesting for later.

tomkerkhove: That way we have:
1) A message pump which checks the entity for new messages and triggers corresponding message handler
2) A message handler who takes action on the deserialized message

src/Arcus.Messaging.Tests.Workers/MessageHandlers/OrdersMessagePump.cs
```diff
@@ -14,32 +14,12 @@ namespace Arcus.Messaging.Tests.Workers.MessageHandlers
 {
     public class OrdersMessagePump : AzureServiceBusMessagePump<Order>
```
tomkerkhove: This should no longer be required

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -18,7 +18,7 @@ namespace Arcus.Messaging.Pumps.ServiceBus
     ///     Message pump for processing messages on an Azure Service Bus entity
     /// </summary>
     /// <typeparam name="TMessage">Type of expected message payload</typeparam>
-    public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessage, AzureServiceBusMessageContext>
+    public class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessage, AzureServiceBusMessageContext>
```
tomkerkhove: `TMessage` should no longer be required since that's what the `IMessageHandler` is for.

`MessagePump` should create an instance of `IMessageHandler` and hand it over to it

stijnmoreels: So, without first deserializing?

tomkerkhove: Ideally, the pump deserializes and passes it on to the message handler based on the type.

Maybe better to do a meeting on this?

stijnmoreels: Updated this, can you take a look if it's something you had in mind?

src/Arcus.Messaging.Tests.Unit/ServiceBus/EmptyMessagePump.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Arcus.Messaging.Tests.Unit.ServiceBus
 {
-    public class EmptyMessagePump : AzureServiceBusMessagePump<Order>
+    public class EmptyMessagePump : AzureServiceBusMessagePump
```
tomkerkhove: In theory you don't need this anymore

src/Arcus.Messaging.Tests.Workers.ServiceBus.Queue/Program.cs
```diff
@@ -36,7 +36,9 @@ public static void Main(string[] args)
                             .UsingAuthenticationKey(eventGridKey)
                             .Build();
                     });
-                    services.AddServiceBusQueueMessagePump<OrdersMessagePump>(configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING"]);
+                    services.AddServiceBusQueueMessagePump(configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING"])
```
tomkerkhove: This syntax looks perfect 👌

src/Arcus.Messaging.Tests.Workers/MessageHandlers/OrdersMessageHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        public async Task<MessageProcessResult> ProcessMessageAsync(
+            Message message, 
+            AzureServiceBusMessageContext messageContext, 
+            MessageCorrelationInfo correlationInfo, 
+            CancellationToken cancellationToken)
+        {
+            Encoding encoding = messageContext.GetMessageEncodingProperty();
+            if (message.TryReadBodyAsJson(encoding, out Order order))
```
tomkerkhove: This should not be the responsibility of the message handler, the message pump should already do this and pass the order itself

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -225,18 +257,16 @@ public static IServiceCollection AddServiceBusTopicMessagePump<TMessagePump>(thi
...
+            AddServiceBusMessagePump<AzureServiceBusMessagePump>(services, topicName, subscriptionName, getConnectionStringFromConfigurationFunc: getConnectionStringFromConfigurationFunc, configureMessagePump: configureMessagePump);
 
             return services;
         }
 
-        private static void AddServiceBusMessagePump<TMessagePump>(IServiceCollection services, string entityName = null, string subscriptionName = null, Func<IConfiguration, string> getConnectionStringFromConfigurationFunc = null, Func<ISecretProvider, Task<string>> getConnectionStringFromSecretFunc = null, Action<AzureServiceBusMessagePumpOptions> configureMessagePump = null)
-            where TMessagePump : class, IHostedService
+        private static void AddServiceBusMessagePump<TMessagePump>(IServiceCollection services, string entityName = null, string subscriptionName = null, Func<IConfiguration, string> getConnectionStringFromConfigurationFunc = null, Func<ISecretProvider, Task<string>> getConnectionStringFromSecretFunc = null, Action<AzureServiceBusMessagePumpOptions> configureMessagePump = null) where TMessagePump : class, IHostedService
```
tomkerkhove: Do we still need the typed pump in the extensions?
```suggestion
        private static void AddServiceBusMessagePump(IServiceCollection services, string entityName = null, string subscriptionName = null, Func<IConfiguration, string> getConnectionStringFromConfigurationFunc = null, Func<ISecretProvider, Task<string>> getConnectionStringFromSecretFunc = null, Action<AzureServiceBusMessagePumpOptions> configureMessagePump = null) where TMessagePump : class, IHostedService
```

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -83,62 +82,80 @@ protected virtual Task HandleReceiveExceptionAsync(Exception receiveException)
...
             {
-                try
+                if (messageHandler is null)
                 {
-                    return Encoding.GetEncoding(annotatedEncoding.ToString());
+                    continue;
                 }
-                catch (Exception ex)
+
+                bool isProcessed = await TryProcessMessageAsync(messageHandler, message, messageContext, correlationInfo, cancellationToken);
```
tomkerkhove: This is tricky, as discussed we should:
1. Try to deserialize the message according to the expected contract of the handler
2. If it matches, only the first match should process it
3. Done

Otherwise we're at risk of running a lot of duplicate processing, which failed, but still can cause inconsistent state

stijnmoreels:
@tomkerkhove , reflection approach available here: https://github.com/stijnmoreels/arcus.messaging/pull/1

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -82,63 +81,71 @@ protected virtual Task HandleReceiveExceptionAsync(Exception receiveException)
             return Task.CompletedTask;
         }
 
+        // TODO: move to extension?
```
tomkerkhove: That's also an option, both are fine for me.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -82,63 +81,71 @@ protected virtual Task HandleReceiveExceptionAsync(Exception receiveException)
...
+                                       $"Unable to determine encoding with name '{{Encoding}}'. Falling back to {{FallbackEncoding}}.",
+                                       annotatedEncoding.ToString(), DefaultEncoding.WebName);
+                }
             }
 
-            return messageBody;
+            return DefaultEncoding;
         }
 
+        // TODO: should we also provide a 'streamed' version to process the message? We would require some way to reset the stream though (i.e. VirtualStream).
```
tomkerkhove: Let's hold this off for now maybe?

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
```diff
@@ -70,5 +72,70 @@ public static string GetTransactionId(this Message message)
...
+        /// <summary>
+        ///     Tries to read the Service Bus message body as a JSON structure, deserialized to the specified <typeparamref name="TMessage"/>.
+        /// </summary>
+        /// <typeparam name="TMessage">The type of the message to be deserialized.</typeparam>
+        /// <param name="message">The message to deserialize.</param>
+        /// <param name="output">The output to where the result should be written.</param>
+        /// <returns>
+        ///     Whether or not the deserialization succeeded [true], or not [false].
+        /// </returns>
+        public static bool TryReadBodyAsJson<TMessage>(this Message message, out TMessage output) where TMessage : class
```
tomkerkhove: Maybe good to add some unit tests for these new methods? They are pretty important.

src/Arcus.Messaging.Tests.Workers/MessageHandlers/OrdersMessageHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+using Arcus.Messaging.Abstractions;
+using Arcus.Messaging.Pumps.ServiceBus;
+using Arcus.Messaging.Tests.Core.Events.v1;
+using Arcus.Messaging.Tests.Core.Messages.v1;
+using GuardNet;
+using Microsoft.Extensions.Logging;
+
+namespace Arcus.Messaging.Tests.Workers.MessageHandlers
+{
+    public class OrdersMessageHandler : IAzureServiceBusMessageHandler<Order>
```
tomkerkhove: Thinking out loud - If I were to not care about the Service Bus message context, can I still use `IMessageHandler<Order>`?

That would be very good to have imo. Reduces lockin.

tomkerkhove: For clarity, I see value for both scenarios

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -246,5 +277,19 @@ private static void AddServiceBusMessagePump<TMessagePump>(IServiceCollection se
...
         }
+
+        /// <summary>
+        ///     Adds a <see cref="IAzureServiceBusMessageHandler{TMessage}"/> implementation to process the messages from Azure Service Bus resources.
+        /// </summary>
+        /// <typeparam name="TMessageHandler">The type of the implementation.</typeparam>
+        /// <typeparam name="TMessage"></typeparam>
+        /// <param name="services">The collection of services to use in the application.</param>
+        public static IServiceCollection WithMessagePumpHandler<TMessageHandler, TMessage>(this IServiceCollection services)
+            where TMessageHandler : class, IAzureServiceBusMessageHandler<TMessage>
```
tomkerkhove: This should not enforce `IAzureServiceBusMessageHandler` but `IMessageHandler` instead, otherwise there is no point in the decouple.

Imagine we add event hubs, I should have the option to use the specific interface for that one or use `IMessageHandler `but with less specific integration such as message context.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -197,6 +213,24 @@ public static IServiceCollection AddServiceBusTopicMessagePump<TMessagePump>(thi
...
+        {
+            Guard.NotNull(services, nameof(services));
+
+            AddServiceBusMessagePump<AzureServiceBusMessagePump>(services, topicName, subscriptionName, getConnectionStringFromSecretFunc: secretProvider => secretProvider.GetRawSecretAsync(secretName), configureMessagePump: configureMessagePump);
+
+            return services;
+        }
+
         /// <summary>
         ///     Adds a message handler to consume messages from Azure Service Bus Topic
```
tomkerkhove: We should update all the `AddServiceBusTopicMessagePump` & `AddServiceBusQueueMessagePump` extensions:

```suggestion
        ///     Adds a message pump to consume messages from Azure Service Bus Topic
```

message handlers are what we add via WithMessagePumpHandler

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -246,5 +277,19 @@ private static void AddServiceBusMessagePump<TMessagePump>(IServiceCollection se
             services.AddSingleton(serviceProvider => new AzureServiceBusMessagePumpSettings(entityName, subscriptionName, getConnectionStringFromConfigurationFunc, getConnectionStringFromSecretFunc, messagePumpOptions, serviceProvider));
             services.AddHostedService<TMessagePump>();
         }
+
+        /// <summary>
+        ///     Adds a <see cref="IAzureServiceBusMessageHandler{TMessage}"/> implementation to process the messages from Azure Service Bus resources.
```
tomkerkhove: Shouldn't be specific for Service Bus
```suggestion
        ///     Adds a message handler to process the messages.
```

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -0,0 +1,203 @@
...
+            object callSiteFactory = engine.GetPropertyValue("CallSiteFactory", BindingFlags.NonPublic | BindingFlags.Instance);
+
+            var descriptorLookup = (System.Collections.IEnumerable) callSiteFactory.GetFieldValue("_descriptorLookup");
+
+            var messageHandlers = new Collection<MessageHandler>();
+            foreach (object lookup in descriptorLookup)
+            {
+                var serviceType = (Type) lookup.GetPropertyValue("Key");
+                
+                // TODO: update to more strict message handler filter.
```
tomkerkhove: Let's add an issue and remove the TODO

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Arcus.Messaging.Abstractions;
+using GuardNet;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Arcus.Messaging.Pumps.Abstractions.MessageHandling
```
tomkerkhove: Can we provide tests for our voodoo part? ⚔😵

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Right, is fixed with merging with `master`.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Yes, right, but for the `AzureMessagePump`, that message pump can only use message handlers with a specific context. So, for that pump, we should only register those kind of implementations.

Maybe we should even change the signature to reflect this. `.WithAzureServiceBusMessageHandler`.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Yes

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: Done: https://github.com/arcus-azure/arcus.messaging/issues/68

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: We should really test the outside of this. Like, registering several message handlers with different messages and see if we can select them all.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: See comment above... this message pump can only handle those kind of message/contexts. We should maybe change the signature.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: We could maybe provide both. Since indeed, we could also register message handlers in a different context. But those will not be selected when processing messages in a Service Bus message pump.
That extension should probably be in another namespace too.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Added a different extension for that.

src/Arcus.Messaging.Tests.Workers/MessageHandlers/OrdersMessageHandler.cs
stijnmoreels: Well, the `IMessageHandler` uses the context in it's signature. So... we would have to create an different `IMessageHandler` interface with maybe an "ignored" message context so we can also include that in the filter.
When processing the message, the context would then be empty. or a `null`-type.

stijnmoreels: In any case; different PR. Can you make an issue for this?

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
stijnmoreels: Yeah, removed that. Leftover from legacy approach.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Yes, of course.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Done that.

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageContextExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            if (messageContext.Properties.TryGetValue(PropertyNames.Encoding, out object annotatedEncoding))
+            {
+                try
+                {
+                    return Encoding.GetEncoding(annotatedEncoding.ToString());
+                }
+                catch
+                {
+                    // Ignore.
```
tomkerkhove: Thought - Should we log it? Because something was specified but we couldn't parse it so we'll use UTF 8 instead.

Migth be easier for troubleshooting

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
tomkerkhove: Question is, why shouldn't we be allowed to use the generic message handlers? if I don't need the service bus message context and regular context is fine, it should be possible, no?

stijnmoreels:
> Question is, why shouldn't we be allowed to use the generic message handlers? if I don't need the service bus message context and regular context is fine, it should be possible, no?

I added another extension method to add just message handlers. The `AzureServiceMessagePump` only can handle handlers in its context, right? Or should we abbandon the context and just send the message with a `null` context instead then?

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageContextExtensions.cs
stijnmoreels: Hmm, then we have to pass in the logger.

tomkerkhove: I wouldn't make it optional, otherwise you are in the dark

src/Arcus.Messaging.Tests.Workers/MessageHandlers/OrdersMessageHandler.cs
tomkerkhove: See other comment, message context will always be there but it will just be specific to SB or the generic model.

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageContextExtensions.cs
tomkerkhove: Nevermind, it's ok

tomkerkhove:
> I added another extension method to add just message handlers. The `AzureServiceMessagePump` only can handle handlers in its context, right? Or should we abbandon the context and just send the message with a `null` context instead then?

It's simple: 
- When I use `IMessageHandler` I will only have `MessageContext`, regardless of the message pump that is being used
- When I use `IServiceBusMessageHandler` I'm explicitly pinning to that broker so I will get `ServiceBusMessagContext`

stijnmoreels:
> > I added another extension method to add just message handlers. The `AzureServiceMessagePump` only can handle handlers in its context, right? Or should we abbandon the context and just send the message with a `null` context instead then?
> 
> It's simple:
> 
> * When I use `IMessageHandler` I will only have `MessageContext`, regardless of the message pump that is being used
> * When I use `IServiceBusMessageHandler` I'm explicitly pinning to that broker so I will get `ServiceBusMessagContext`

Updated this with another shortcut message handler.

stijnmoreels:
> LGTM - Do you want to do docs here or seperate PR?

Yes, I think that would be best.

PR #50, opened 2020-01-27
FEAT: add Azure ServiceBus Queue/Topic MessagePump registration shortcuts

### LINKED ISSUE #48: Provide simplified approach to wire up the Service Bus pump
Provide simplified version to use Service Bus pump which does not require you to use a func to get the connection string, but a secret key name instead:
```csharp
services.AddServiceBusTopicMessagePump<EmptyMessagePump>("Receive-All", "My-Secret-Name");
```

The documentation should clearly state that it will look it up by using `ISecretProvider` and not `IConfigurationProvider`.

stijnmoreels:
Closes #48

### DIFF
--- a/src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
+++ b/src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
@@ -1,0 +1,53 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Arcus.Messaging.Tests.Unit.ServiceBus
+{
+    public class IServiceCollectionExtensionsTests
+    {
+        [Fact]
+        public void AddServiceBusTopicMessagePump_WithSubscriptionNameIndirectSecretProvider_WiresUpCorrectly()
+        {
+            // Arrange
+            var services = new ServiceCollection();
+
+            // Act
+            IServiceCollection result = 
+                services.AddServiceBusTopicMessagePump<EmptyMessagePump>(
+                    "subscription name", "secret name", configureMessagePump: options => options.AutoComplete = true);
+            
+            // Assert
+            Assert.NotNull(result);
+        }
+
+        [Fact]
+        public void AddServiceBusTopicMessagePump_WithTopicNameAndSubscriptionNameIndirectSecretProvider_WiresUpCorrectly()
+        {
+            // Arrange
+            var services = new ServiceCollection();
+
+            // Act
+            IServiceCollection result = 
+                services.AddServiceBusTopicMessagePump<EmptyMessagePump>(
+                    "topic name", "subscription name", "secret name", configureMessagePump: options => options.AutoComplete = true);
+            
+            // Assert
+            Assert.NotNull(result);
+        }
+
+        [Fact]
+        public void AddServiceBusQueueMessagePump_IndirectSecretProvider_WiresUpCorrectly()
+        {
+            // Arrange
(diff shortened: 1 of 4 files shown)


tomkerkhove:
LGTM, one small suggestion would be to maybe add a quick test if easy to do.

stijnmoreels:
> LGTM, one small suggestion would be to maybe add a quick test if easy to do.

Well, that was indeed something I was thinking about. Most of the extension methods in that file are never used. (except for the ones used in the dummy test projects). So, unless I add another project to verify this extension method?

tomkerkhove:
We should provide some unit tests for them imo, just to see if they are wired up all fine or not - But ok to do via seperate issue if you want.

stijnmoreels:
> We should provide some unit tests for them imo, just to see if they are wired up all fine or not - But ok to do via seperate issue if you want.

Ah, yes. That could be done.

src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,53 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Arcus.Messaging.Tests.Unit.ServiceBus
```
tomkerkhove: Shouldn't we also assert that it registers the correct lookup?

stijnmoreels: Hmm, how would you do that without making it an integration test?

tomkerkhove: Check the type that was registered in the collection and how it looks it up, if possible.

Or use mocking to see which one was called how many times?

I don't know, just thinking out loud.

stijnmoreels: Ah, now I see what you mean. I'll check that.

src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+{
+    public class IServiceCollectionExtensionsTests
+    {
+        [Fact]
+        public async Task AddServiceBusTopicMessagePump_WithSubscriptionNameIndirectSecretProvider_WiresUpCorrectly()
+        {
+            // Arrange
+            var services = new ServiceCollection();
+            var spySecretProvider = new Mock<ISecretProvider>();
```
tomkerkhove: Spy 😂

tomkerkhove:
Would be great to just update the docs as well, thanks!

src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
stijnmoreels: 😄 yeah, believe it or not; but that's the official terminology of xUnit (not the library but the pattern) testing practices 😃 
http://xunitpatterns.com/Test%20Spy.html

stijnmoreels:
> Would be great to just update the docs as well, thanks!

Do I have this right? There's not docs for anything message pump related? That should maybe even has it's own issue/PR bc it's propably much.

tomkerkhove:
Oh yeah my bad, that's actually on my list 😂

Issue #39, opened 2019-12-18
Provide option to create Service Bus Topic subscription if it doesn't exist


tomkerkhove:
Provide option to create Service Bus Topic subscription if it doesn't exist.

This should be done while registration of the message pump by using `AzureServiceBusMessagePumpOptions` and document the required permissions, most probably `Manage`.

tomkerkhove:
#62 relates to this

PR #52, opened 2020-02-04
FEAT: add topic subscription option to message pump

stijnmoreels:
Add an option to include the creation/deletion Auzre Service Bus Topic subscription during the start/stop of the message pump.

#39


stijnmoreels:
Was it something like this you had in mind?

tomkerkhove:
Had a quick look and approach looks good, but will provide better thorough review soon.

Would you mind updating the docs as well please?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -35,7 +37,9 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
             : base(configuration, serviceProvider, logger)
         {
             Settings = serviceProvider.GetRequiredService<AzureServiceBusMessagePumpSettings>();
+            JobId = Settings.Options.JobId;
 
+            _subscriptionName = $"{Settings.SubscriptionPrefix}:{JobId}";
```
tomkerkhove: Let's use a hyphen instead, this is more commonly used
```suggestion
            _subscriptionName = $"{Settings.SubscriptionPrefix}-{JobId}";
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +53,42 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+
+        /// <summary>
+        /// Triggered when the application host is ready to start the service.
+        /// </summary>
+        /// <param name="cancellationToken">Indicates that the start process has been aborted.</param>
+        public override async Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (Settings.Options.IncludeTopicSubscription)
+            {
+                ServiceBusConnectionStringBuilder serviceBusConnectionString = await GetServiceBusConnectionStringAsync();
```
tomkerkhove: Let's move all of this to a dedicated method

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +53,42 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+
+                Logger.LogTrace("[Job: {JobId}] Creating subscription '{SubscriptionName}' on topic '{TopicPath}'...", JobId, _subscriptionName, serviceBusConnectionString.EntityPath);
```
tomkerkhove: Do we use the `[Job: {JobId}] ` prefix for other traces as well? If not, please align this line or the others

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +53,42 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+                var subscriptionDescription = new SubscriptionDescription(serviceBusConnectionString.EntityPath, _subscriptionName)
+                {
+                    AutoDeleteOnIdle = TimeSpan.FromHours(1),
+                    MaxDeliveryCount = 3,
+                    UserMetadata = $"Subscription created by Arcus job: '{JobId}' to process inbound CloudEvents."
```
tomkerkhove: Copy paste?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +53,42 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+        public override async Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (Settings.Options.IncludeTopicSubscription)
+            {
+                ServiceBusConnectionStringBuilder serviceBusConnectionString = await GetServiceBusConnectionStringAsync();
+
+                Logger.LogTrace("[Job: {JobId}] Creating subscription '{SubscriptionName}' on topic '{TopicPath}'...", JobId, _subscriptionName, serviceBusConnectionString.EntityPath);
+                var subscriptionDescription = new SubscriptionDescription(serviceBusConnectionString.EntityPath, _subscriptionName)
+                {
+                    AutoDeleteOnIdle = TimeSpan.FromHours(1),
```
tomkerkhove: Let's not set this one so that ops can troubleshoot things

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +53,42 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+                    MaxDeliveryCount = 3,
```
tomkerkhove: Not really up to us to define so I'd suggest to leave this out for now

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +53,42 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+                    UserMetadata = $"Subscription created by Arcus job: '{JobId}' to process inbound CloudEvents."
+                };
+            
+                var ruleDescription = new RuleDescription("Accept-All", new TrueFilter());
+
+                var serviceBusClient = new ManagementClient(serviceBusConnectionString);
+                await serviceBusClient.CreateSubscriptionAsync(subscriptionDescription, ruleDescription, cancellationToken)
+                                      .ConfigureAwait(continueOnCapturedContext: false);
```
tomkerkhove: We don't do this in other places nor it's a library, maybe best to remove for now then or not?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +53,42 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+
+                Logger.LogTrace("[Job: {JobId}] Subscription '{SubscriptionName}' created on topic '{TopicPath}'", JobId, _subscriptionName, serviceBusConnectionString.EntityPath);
+                await serviceBusClient.CloseAsync().ConfigureAwait(continueOnCapturedContext: false);
```
tomkerkhove: We don't do this in other places nor it's a library, maybe best to remove for now then or not?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +53,42 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+                    UserMetadata = $"Subscription created by Arcus job: '{JobId}' to process inbound CloudEvents."
+                };
+            
+                var ruleDescription = new RuleDescription("Accept-All", new TrueFilter());
+
+                var serviceBusClient = new ManagementClient(serviceBusConnectionString);
+                await serviceBusClient.CreateSubscriptionAsync(subscriptionDescription, ruleDescription, cancellationToken)
+                                      .ConfigureAwait(continueOnCapturedContext: false);
+
+                Logger.LogTrace("[Job: {JobId}] Subscription '{SubscriptionName}' created on topic '{TopicPath}'", JobId, _subscriptionName, serviceBusConnectionString.EntityPath);
```
tomkerkhove: Same here about aligning

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
tomkerkhove: Let's make it a property instead

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -250,10 +290,31 @@ protected virtual IEnumerable<ServiceBusPlugin> DefineServiceBusPlugins()
         /// <param name="cancellationToken">Indicates that the shutdown process should no longer be graceful.</param>
         public override async Task StopAsync(CancellationToken cancellationToken)
         {
+            if (Settings.Options.IncludeTopicSubscription)
```
tomkerkhove: Creating is one thing, deleting is another.

I'd suggest to leave this out or do this via another config option.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -34,9 +34,28 @@ public class AzureServiceBusMessagePumpOptions
...
 
+        /// <summary>
+        /// Gets or sets the unique identifier for this background job to distinguish this job instance in a multi-instance deployment.
+        /// </summary>
+        public string JobId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value indicating whether or not a new Azure Service Bus Topic subscription has to be created when the <see cref="AzureServiceBusMessagePump{TMessage}"/> starts.
+        /// The subscription will be deleted afterwards when the message pump stops.
+        /// REMARK: requires at least 'Manage' permissions on the Azure Service Bus Topic to create and delete these subscriptions.
```
tomkerkhove: Can be removed since it's in the `remarks` section
```suggestion
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -34,9 +34,28 @@ public class AzureServiceBusMessagePumpOptions
...
+        /// </summary>
+        /// <remarks>
+        ///     Requires at least 'Manage' permissions on the Azure Service Bus Topic to create and delete these subscriptions.
```
tomkerkhove: Manage is the highest permission
```suggestion
        ///     Requires 'Manage' permissions on the Azure Service Bus Topic to create and delete these subscriptions.
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -34,9 +34,28 @@ public class AzureServiceBusMessagePumpOptions
...
         public bool AutoComplete { get; set; }
 
+        /// <summary>
+        /// Gets or sets the unique identifier for this background job to distinguish this job instance in a multi-instance deployment.
+        /// </summary>
+        public string JobId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value indicating whether or not a new Azure Service Bus Topic subscription has to be created when the <see cref="AzureServiceBusMessagePump{TMessage}"/> starts.
+        /// The subscription will be deleted afterwards when the message pump stops.
```
tomkerkhove: See other comment - I'd do thise via seperate option or not at all

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpSettings.cs
```diff
@@ -22,12 +22,12 @@ public class AzureServiceBusMessagePumpSettings
         ///     Constructor
         /// </summary>
         /// <param name="entityName">Name of the entity to process</param>
-        /// <param name="subscriptionName">Name of the subscription to process</param>
+        /// <param name="subscriptionPrefix">Name of the subscription to process</param>
```
tomkerkhove:
```suggestion
        /// <param name="subscriptionPrefix">Name prefix of the subscription to process</param>
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -34,9 +34,28 @@ public class AzureServiceBusMessagePumpOptions
...
+        /// REMARK: requires at least 'Manage' permissions on the Azure Service Bus Topic to create and delete these subscriptions.
+        /// </summary>
+        /// <remarks>
+        ///     Requires at least 'Manage' permissions on the Azure Service Bus Topic to create and delete these subscriptions.
+        /// </remarks>
+        public bool IncludeTopicSubscription { get; set; }
```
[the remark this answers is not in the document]
stijnmoreels: Yeah, but I also wanted to have in the name something that would 'destroy' the subscription too, you know?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Yes.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Yes.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Woops.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Ok.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: I guess, the creation/deletion of topic subscription can be handled independent of the actual message pump; that's why I've left it here.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Ok.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: For all log traces, yes, but not all log messages. Like information messages doesn't have this. Do you want me to align it to every log entry that's being written in the pump?

tomkerkhove: Let's do that, but in a follow-up PR

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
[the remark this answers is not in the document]
stijnmoreels: Yep, updated with an enum instead.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Ref: https://github.com/arcus-azure/arcus.messaging/issues/58

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -46,15 +46,16 @@ public class AzureServiceBusMessagePumpOptions
...
 
         /// <summary>
         ///     Default settings
         /// </summary>
         internal static AzureServiceBusMessagePumpOptions Default => new AzureServiceBusMessagePumpOptions
         {
             AutoComplete = true,
-            JobId = Guid.NewGuid().ToString()
+            JobId = Guid.NewGuid().ToString(),
+            TopicSubscription = TopicSubscription.None
```
tomkerkhove: I would turn on subscription creation by default

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -46,15 +46,16 @@ public class AzureServiceBusMessagePumpOptions
         /// <remarks>
         ///     Requires 'Manage' permissions on the Azure Service Bus Topic to create and delete these subscriptions.
```
tomkerkhove: How about this?
```suggestion
        ///     Provides capability to create and delete these subscriptions. This requires 'Manage' permissions on the Azure Service Bus Topic or namespace.
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +52,55 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+        {
+            if ((Settings.Options.TopicSubscription & TopicSubscription.CreateOnStart) == TopicSubscription.CreateOnStart)
+            {
+                await CreateTopicSubscriptionAsync(cancellationToken);
+            }
+
+            await base.StartAsync(cancellationToken);
+        }
+
+        private async Task CreateTopicSubscriptionAsync(CancellationToken cancellationToken)
```
tomkerkhove: Thinking out loud - We could move this to a `AzureServiceBusTopicManager` or so, but not a must.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -250,10 +302,43 @@ protected virtual IEnumerable<ServiceBusPlugin> DefineServiceBusPlugins()
...
+            if ((Settings.Options.TopicSubscription & TopicSubscription.DeleteOnStop) == TopicSubscription.DeleteOnStop)
+            {
+                await DeleteTopicSubscriptionAsync(cancellationToken);
+            }
+
             await base.StopAsync(cancellationToken);
             _isHostShuttingDown = true;
         }
 
+        private async Task DeleteTopicSubscriptionAsync(CancellationToken cancellationToken)
```
tomkerkhove: Thinking out loud - We could move this to a `AzureServiceBusTopicManager` or so, but not a must.

tomkerkhove: If so, a follow-up PR for this might be better maybe 🤔

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -250,10 +302,43 @@ protected virtual IEnumerable<ServiceBusPlugin> DefineServiceBusPlugins()
         /// <param name="cancellationToken">Indicates that the shutdown process should no longer be graceful.</param>
         public override async Task StopAsync(CancellationToken cancellationToken)
         {
+            if ((Settings.Options.TopicSubscription & TopicSubscription.DeleteOnStop) == TopicSubscription.DeleteOnStop)
```
tomkerkhove: Is this a shorthand notation of `Settings.Options.TopicSubscription != null && Settings.Options.TopicSubscription == TopicSubscription.DeleteOnStop`?

stijnmoreels:
I actually don't understand why the integration tests fails. It looks like a configuration issue rather than implementation. :(

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: The enum is actually just a number (so can't be `null`). I added the `[Flags]` attribute to indicate that we expect combinations of the enum values.

So, the notation is a bitwise condition to filter out first the `.DeleteOnStop` value (which is `3` from the configured options value (`1` for example).
Bitwise is `3`, `0011` and `1`, `0001`. The `&` filters out all other bits. Like if we would have set the options to `7` (`0111`) it would become `0011`.
This final value (`0011`) we can use to match against the thing we actually want to check `3`.

So, in short, the syntax allows you to first filter out ("clean" if you will) the bits that would make up the expected value, and then check if it's indeed the expected value.
You follow?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: If it would delete the topic on dispose, I would make it a `IDisposable`; but that's not the case. So, hmm, maybe just too little functionality to get the added value to move this?

stijnmoreels:
> Docs will be via a dedicated PR I presume or think it's not worth it?

I can add the docs here. But yeah, there's not much to document 😄 .

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
stijnmoreels: Creation but not deletion?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
tomkerkhove: Would be seperation of concerns but it's ok

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
tomkerkhove: Can't we just use `Settings.Options.TopicSubscription.HasFlag(TopicSubscription.DeleteOnStop)` which does what you want to achieve?

tomkerkhove:
> I actually don't understand why the integration tests fails. It looks like a configuration issue rather than implementation. :(

I'll have a look, maybe it's because it doesn't have management permissions; but if that is the case it would be ideal to provide good logs around that (maybe local testing is better to figure this out though)

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Ah, yes, that would make the code maybe more expressive, but the really downside on that approach is that it has a unnecessary boxing/unboxing on the integer bc it also checks the type of the enum. Bitwise operations are much faster here.
So, ... IDK. I can change it if you like. There's not that much to win I guess.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
tomkerkhove: Jup

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
tomkerkhove: Let's just follow the typical .NET approach by using `HasFlag`

tomkerkhove:
https://github.com/arcus-azure/arcus.messaging/pull/59 should give you insights on the Docker container when tests fail

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
stijnmoreels: Yeah, is OK, I just alwasy think in a "clean after you're done"-approach 🗑

stijnmoreels:
> #59 should give you insights on the Docker container when tests fail

I think it's about the conenction string of the Service Bus topic? And that we have to set it to 'Manage'? 😝 But I find it strange, cause there's not a single shared access policy configured on the Service Bus Topic.... What is it using now, then? I wonder.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
tomkerkhove: Let's do that indeed, better

tomkerkhove:
Looks like it's having issues with creating the subscription:
>  ---> System.ArgumentException: The specified HTTP verb (PUT) is not valid. To know more visit https://aka.ms/sbResourceMgrExceptions. . TrackingId:3c110d46-afef-4833-8574-f20f5834fd2c_G20, SystemTracker:arcus-messaging-dev-we-integration-tests:Queue:orders, Timestamp:2020-02-10T12:31:35
   --- End of inner exception stack trace ---
   at Microsoft.Azure.ServiceBus.Management.ManagementClient.SendHttpRequest(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Azure.ServiceBus.Management.ManagementClient.PutEntity(String path, String requestBody, Boolean isUpdate, String forwardTo, String fwdDeadLetterTo, CancellationToken cancellationToken)
   at Microsoft.Azure.ServiceBus.Management.ManagementClient.CreateSubscriptionAsync(SubscriptionDescription subscriptionDescription, RuleDescription defaultRule, CancellationToken cancellationToken)

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +52,55 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+                JobId, SubscriptionName, serviceBusConnectionString.EntityPath);
+            
+            var subscriptionDescription = new SubscriptionDescription(serviceBusConnectionString.EntityPath, SubscriptionName)
+            {
+                UserMetadata = $"Subscription created by Arcus job: '{JobId}' to process Service Bus messages."
+            };
+
+            var ruleDescription = new RuleDescription("Accept-All", new TrueFilter());
+            var serviceBusClient = new ManagementClient(serviceBusConnectionString);
+            await serviceBusClient.CreateSubscriptionAsync(subscriptionDescription, ruleDescription, cancellationToken)
```
tomkerkhove: YOu shuld check if the subscription with same name already exists

stijnmoreels: Yep done. Thanks.

stijnmoreels:
@tomkerkhove Ok, all fixed, the problem was that their was that the job ID was set to the subscription name in a Queue message pump when in that case the subscription name should be `null`.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -49,6 +60,66 @@ public abstract class AzureServiceBusMessagePump<TMessage> : MessagePump<TMessag
...
+                UserMetadata = $"Subscription created by Arcus job: '{JobId}' to process Service Bus messages."
+            };
+
+            var ruleDescription = new RuleDescription("Accept-All", new TrueFilter());
+            var serviceBusClient = new ManagementClient(serviceBusConnectionString);
+
+            bool subscriptionExists = 
+                await serviceBusClient.SubscriptionExistsAsync(
+                    serviceBusConnectionString.EntityPath, SubscriptionName, cancellationToken);
+            if (subscriptionExists)
```
tomkerkhove: Let's do this before we even start composing the rule description, log it as you have it now and return. No need for an else then

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -34,9 +34,28 @@ public class AzureServiceBusMessagePumpOptions
...
+
         /// <summary>
         ///     Default settings
         /// </summary>
-        internal static AzureServiceBusMessagePumpOptions Default => new AzureServiceBusMessagePumpOptions { AutoComplete = true };
+        internal static AzureServiceBusMessagePumpOptions Default => new AzureServiceBusMessagePumpOptions
+        {
+            AutoComplete = true,
+            JobId = Guid.NewGuid().ToString(),
+            TopicSubscription = TopicSubscription.CreateOnStart
```
tomkerkhove: Didn't we decide to clean up after finished like you've suggested? If so, please don't forget to update docs with new default

stijnmoreels: I suggested it, but wasn't sure you agreed 😄 .

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: I'll move the rule description, use a try/finaly block for the service bus client to be closed, but I don't agree that removing the else and adding a return statement will benifit readability. If the return statement is forgotten, or removed by mistake, we got invalid code, while with the else this is much clearer.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
tomkerkhove: It's a fair suggestion so go ahead! 👍

docs/features/message-pumps/service-bus.md
```diff
@@ -106,8 +106,8 @@ public void ConfigureServices(IServiceCollection services)
             options.JobId = Guid.NewGuid().ToString();
 
             // Indicate whether or not a new Azure Service Bus Topic subscription should be created/deleted
-            // when the message pump starts/stops (default: CreateOnStart).
-            options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
+            // when the message pump starts/stops (default: DeleteOnStop, which implicitly means also 'create on start').
+            options.TopicSubscription = TopicSubscription.DeleteOnStop;
```
tomkerkhove: Ha think there is some miscommunication, I'd keep both by default.
```suggestion
            options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -55,7 +55,7 @@ public class AzureServiceBusMessagePumpOptions
         {
             AutoComplete = true,
             JobId = Guid.NewGuid().ToString(),
-            TopicSubscription = TopicSubscription.CreateOnStart
+            TopicSubscription = TopicSubscription.DeleteOnStop
```
tomkerkhove:
```suggestion
            TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
```

docs/features/message-pumps/service-bus.md
stijnmoreels: Yes, I know 😄 . `CreateOnStart = 1` and `DeleteOnStop = 3` so bitwise, if you choose `DeleteOnStop` you implicitly say that you want to create it on start. So it's the same.

stijnmoreels: But in the docs, we could make explicit indeed. In code, not very much added value I think.

tomkerkhove: Frankly, it should not be that way, how would I then delete it on stop without creating it here if I want?

stijnmoreels: It's implicit saying that you will create it at start. It's actually kind of a failsafe. So, that you can't configure the pump with only deleting the subscription, without also implicitly stating that you want it to be created at startup.
For readability, I can understand that it makes sense to use them both, but behind the scenes, there's no difference.

tomkerkhove: I don't like hidding things, we should just use a flag approach and if only `DeleteOnStop` is specified only do that. If the subscription is not there, throw an exception, it's their fault since they've configured it that way 🤷‍♂️

But default should be `CreateOnStart | DeleteOnStop` indeed.

tomkerkhove: Reason is - Sometimes subscriptions are created externally so we should not implicitly do this if it's not specified.

docs/features/message-pumps/service-bus.md
```diff
@@ -57,25 +57,64 @@ public void ConfigureServices(IServiceCollection services)
...
+            // The amount of concurrent calls to process messages 
+            // (default: null, leading to the defaults of the Azure Service Bus SDK message handler options).
+            options.MaxConcurrentCalls = 5;
+
+            // The unique identifier for this background job to distinguish 
+            // this job instance in a multi-instance deployment (default: guid).
+            options.JobId = Guid.NewGuid().ToString();
+
+            // Indicate whether or not a new Azure Service Bus Topic subscription should be created/deleted
+            // when the message pump starts/stops (default: DeleteOnStop, which implicitly means also 'create on start').
```
tomkerkhove:
```suggestion
            // when the message pump starts/stops (default: CreateOnStart & DeleteOnStop).
```

tomkerkhove:
Thanks!

Issue #22, opened 2019-12-04
Provide Azure Service Bus plugin for adding correlation info to messages


tomkerkhove:
Provide Azure Service Bus plugin for adding correlation info to outbound messages based on the current `ICorrelationInfo`.

For an example of what a Service Bus plugin is, see [SeanFeldman/ServiceBus.AttachmentPlugin](https://github.com/SeanFeldman/ServiceBus.AttachmentPlugin).

stijnmoreels:
I think we can use the ServiceBus plugin functionality to add correlation information:

```csharp
/// <summary>
    /// Azure ServiceBus plugin to add correlation information to the outbound message.
    /// </summary>
    public class CorrelationInfoServiceBusPlugin : ServiceBusPlugin
    {
        private readonly CorrelationInfo _correlationInfo;

        /// <summary>
        /// Initializes a new instance of the <see cref="CorrelationInfoServiceBusPlugin"/> class.
        /// </summary>
        public CorrelationInfoServiceBusPlugin(CorrelationInfo correlationInfo)
        {
            Guard.NotNull(correlationInfo, nameof(correlationInfo));

            _correlationInfo = correlationInfo;
        }

        /// <inheritdoc />
        public override string Name => nameof(CorrelationInfoServiceBusPlugin);

        /// <inheritdoc />
        public override Task<Message> BeforeMessageSend(Message message)
        {
            message.CorrelationId = _correlationInfo.OperationId;

            if (!String.IsNullOrWhiteSpace(_correlationInfo.TransactionId))
            {
                message.UserProperties.Add(PropertyNames.TransactionId, _correlationInfo.TransactionId);
            }

            return Task.FromResult(message);
        }
    }
```

stijnmoreels:
Can we use auto DI container instead registering manually.

stijnmoreels:
Is this what we look for:
https://github.com/SeanFeldman/ServiceBus.AttachmentPlugin/blob/develop/src/ServiceBus.AttachmentPlugin/AzureStorageAttachmentExtensions.cs

tomkerkhove:
> Is this what we look for:
[...]

PR #53, opened 2020-02-04
WIP: FEAT: add Service Bus plugin to enrich outgoing messages with correlation information

stijnmoreels:
Relates to #22

### DIFF
--- a/src/Arcus.Messaging.Pumps.ServiceBus/OutboundCorrelationInfoServiceBusPlugin.cs
+++ b/src/Arcus.Messaging.Pumps.ServiceBus/OutboundCorrelationInfoServiceBusPlugin.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Threading.Tasks;
+using Arcus.Messaging.Abstractions;
+using GuardNet;
+using Microsoft.Azure.ServiceBus;
+using Microsoft.Azure.ServiceBus.Core;
+
+namespace Arcus.Messaging.Pumps.ServiceBus
+{
+    /// <summary>
+    /// Plugin to provide correlation information at outgoing Azure Service Bus messages.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


stijnmoreels:
Just pushed this bc was not sure what it was again that we would need for this one.
I guess it's not entirely correct?

tomkerkhove:
Approach is good, but we cannot inject the correlation info as it's scoped to a single message. Since the plugin is only registered once it should dynamically get the required correlation info on the fly

tomkerkhove:
Or if we don't like that; we can provide an accessor that abstracts this for us but should be fetched every time.

stijnmoreels:
> Approach is good, but we cannot inject the correlation info as it's scoped to a single message. Since the plugin is only registered once it should dynamically get the required correlation info on the fly

From where? The pump? Should we register the correlation in the services? But then what about parallel processing of messages...

tomkerkhove:
We should fetch it from the DI container since we register it there but I might be mistaken with Web API.

Bottom line is, how it is now it will not pass the correct correlation info.

stijnmoreels:
> We should fetch it from the DI container since we register it there but I might be mistaken with Web API.
> 
> Bottom line is, how it is now it will not pass the correct correlation info.

I think we better check this offline bc I'm not sure if I get it completly.

tomkerkhove:
What I was referring to was a plugin that works similar to our recent enricher: https://github.com/arcus-azure/arcus.webapi/blob/fdabfcec041afafc8784d6abe9a2507c5e66c781/src/Arcus.WebApi.Telemetry.Serilog/Correlation/CorrelationInfoEnricher.cs

Services are passed, correlation is retrieved on a per-message basis and done. Unless we don't register correlation info in messaging yet of course.

PR #56, opened 2020-02-05
Provide overload that doesn't require queue name

tomkerkhove:
Provide overload that doesn't require queue name, only secret name.

This is to support connection strings scoped to the queue itself.

Follow-up of #50


stijnmoreels:
> Provide overload that doesn't require queue name, only secret name.
> 
> This is to support connection strings scoped to the queue itself.
> 
> Follow-up of #50

So, for topics this is already possible?

tomkerkhove:
Yes, it was added via #50 ([change](https://github.com/arcus-azure/arcus.messaging/pull/50/files#diff-e94dca946f203e80c8cd694175a403feR115))

PR #64, opened 2020-02-22
FIX: reuse subscription across topic message pumps

### LINKED ISSUE #63: `AddServiceBusTopicMessagePump` should require subscription name, not prefix
`AddServiceBusTopicMessagePump` should require subscription name since you'd always want to specify the subscription to process. If the app runs multiple instances, they have to still process the same subscription regardless of the job.

The only exception are things like the Key Vault invalidation job which requires a prefix. For those, I would introduce `AddServiceBusTopicMessagePumpWithPrefix` or so.

This is a blocker for v0.1 imo as current approach is not usable.

stijnmoreels:
Closes #63


stijnmoreels:
> LGTM, can we also update docs please?

Ah, shoot, always forget that. Thanks!

Issue #62, opened 2020-02-20
`TopicSubscription` should not be a configuration option for queues


tomkerkhove:
`TopicSubscription` should not be a configuration option for queues

PR #65, opened 2020-02-25
FIX: provide separate configuration Service Bus message pump options

stijnmoreels:
Relates to #62


stijnmoreels:
There's gonna be maybe a lot of merge conflicts once we introduce the message handling, but yeah. Shit happens 💩 .

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptionsBase.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using GuardNet;
+
+namespace Arcus.Messaging.Pumps.ServiceBus.Configuration 
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class AzureServiceBusMessagePumpOptionsBase
```
tomkerkhove: What is the benefit of not adding this to `AzureServiceBusMessagePumpOptions`?

I don't like Base suffix, it's discouraged by .NET design guidelines as well
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/base-classes-for-implementing-abstractions

tomkerkhove: I think it's because the other one is used by pump internally. Then I would name this AzureServiceBusPumpOptions or whatever

stijnmoreels: "ConfigurationOptions"...

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpConfiguration.cs
```diff
@@ -1,43 +1,42 @@
...
-            set
-            {
-                if (value != null)
-                {
-                    Guard.For<ArgumentException>(() => value <= 0, "Max concurrent calls has to be 1 or above.");
-                }
-
-                _maxConcurrentCalls = value;
-            }
+            Guard.NotNull(options, nameof(options));
```
tomkerkhove: Should we guard on:
- jobId being empty or whatever
- max concurrent calls < 0 (might just rely on the SDK for this though)

stijnmoreels:
> Looks good but confused why we have AzureServiceBusMessagePumpConfiguration, AzureServiceBusMessagePumpOptions & AzureServiceBusMessagePumpSettings

Yeah, I was initially also confused about the Settings <> Options. Apparently there's an distinction between values that you configure that manipulate the functionality of the pump, and ones that don't.

So, `AzureServiceBusMessagePumpSettings` are the final set of values we sent to the message pump.
It is created from either an `...TopicOptions` or `...QueueOptions` model that both inherit from the `...Options` to reduce duplication; that is wrapped in a `...Configuration` model to have a common type.
That last part is bc some options are either  for Topic or Queue.

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpConfiguration.cs
stijnmoreels: Hmmm, that is already check while we create the Topic/Queue options...

tomkerkhove:
Can you talk about how AzureServiceBusMessagePumpSettings and AzureServiceBusMessagePumpConfiugration are different?

stijnmoreels:
> Can you talk about how AzureServiceBusMessagePumpSettings and AzureServiceBusMessagePumpConfiugration are different?

The `...Settings` also contains things like `SubscriptionName`, `GetConnectionStringAsync()` while the `...Configuration` is a model that's created from either a Topic or Queue setup and that the `...Settings` exposes as options that influence the behavior of the message pump.

tomkerkhove:
Last check - Did you verify if docs are up to date?

stijnmoreels:
> Last check - Did you verify if docs are up to date?

They are now 😆 .

PR #69, opened 2020-03-02
Chore - update Messaging projects with netcore3.0 FrameworkReference

stijnmoreels:
Relates to https://github.com/arcus-azure/arcus/issues/50

### DIFF
--- a/src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
+++ b/src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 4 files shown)


stijnmoreels:
Hmm, the `Arcus.Messaging.ServiceBus.Core` references a project that targets both netstandard and netcore. But the project itself only targets netstandard. Is that good?

tomkerkhove:
Normally we should have enough with the netstandard reference, can you clean that up and see if everything still works please?

stijnmoreels:
>> Hmm, the Arcus.Messaging.ServiceBus.Core references a project that targets both netstandard and netcore. But the project itself only targets netstandard. Is that good?
>
> Normally we should have enough with the netstandard reference, can you clean that up and see if everything still works please?

What do you mean with "clean up"? I didn't change the project target framework. So it's already just referencing netstandard.

tomkerkhove:
Ah misread, then we are good to go.

Issue #23, opened 2019-12-04
Provide Azure Service Bus plugin for interpreting correlation info from messages


tomkerkhove:
Provide Azure Service Bus plugin for interpreting correlation info for inbound messages and populate `ICorrelationInfo`, if possible.

This is just an idea and should first be analysed if it's possible and worth the effort.

For an example of what a Service Bus plugin is, see [SeanFeldman/ServiceBus.AttachmentPlugin](https://github.com/SeanFeldman/ServiceBus.AttachmentPlugin).

stijnmoreels:
This is done, no? We introduced an extension for this.

tomkerkhove:
Can you link the PR that added it? Looks like it's not linked here and want to verify that everything is there including docs.

stijnmoreels:
> Can you link the PR that added it? Looks like it's not linked here and want to verify that everything is there including docs.

https://github.com/arcus-azure/arcus.messaging/pull/51

tomkerkhove:
We are missing docs on this, can you provide them please?

PR #71, opened 2020-03-03
Docs - add feature docs for message correlation

stijnmoreels:
Relates to #https://github.com/arcus-azure/arcus.messaging/issues/23


tomkerkhove:
One question though, is that object specific for Service Bus, I don't think so 🤔

stijnmoreels:
> One question though, is that object specific for Service Bus, I don't think so 🤔

It's a `Microsoft.Azure.ServiceBus.Message` object so I'ld say yes.

tomkerkhove:
Good point!

PR #74, opened 2020-03-06
Feature - stricten MessageHandler filter to substract 'IMessageHandler<,>' implementations

### LINKED ISSUE #68: Stricten the `MessageHandler` filtering with a more robust open interface filter
**What whould you like to change?**
Currently, the `MessageHandler` internal type uses only the string `"IMessageHandler"` to find registered types that can be used for message handling. This should be more stricten so there's no way other types of interfaces are selected (`IMessageHandlerSomethingElse` for example).

This filter should take into account that the `IMessageHandler` is an open interface, the namespace, and the single generic type.

stijnmoreels:
Closes #68


stijnmoreels:
Waiting for #75

PR #76, opened 2020-03-07
Chore - move build variables to variable template

stijnmoreels:
Relates to https://github.com/arcus-azure/arcus/issues/46


tomkerkhove:
Why was this closed?

stijnmoreels:
> Why was this closed?

Had some problems with building and am fixing that currently. Still something with the service bus connection strings and/or assertion with event grid.
So, after that, when the build is back green, we can again push this.

Issue #58, opened 2020-02-09
Improvement - Add 'JobId' to every written log entry in the Azure message pump


stijnmoreels:
**What would you like to improve?**
Now, only the log traces in the Azure message pump are enriched with the Job identifier, while each log entry should be enrichted with this.

PR #79, opened 2020-03-10
Feature - add 'JobId' to log messages in the Azure Service Bus message pump

stijnmoreels:
Closes https://github.com/arcus-azure/arcus.messaging/issues/58

### DIFF
--- a/src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
+++ b/src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
@@ -23,9 +23,12 @@ namespace Arcus.Messaging.Pumps.ServiceBus
     /// </summary>
     public class AzureServiceBusMessagePump : MessagePump
     {
+        private readonly MessageHandlerOptions _messageHandlerOptions;
+        private readonly IDisposable _loggingScope;
+        
         private bool _isHostShuttingDown;
         private MessageReceiver _messageReceiver;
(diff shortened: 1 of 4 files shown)


tomkerkhove:
We should have a look if .NET Core allows us to define it once and add it to all following messages.

stijnmoreels:
> We should have a look if .NET Core allows us to define it once and add it to all following messages.

We could use instead an `EventId` for this, but then the `JobId` should be an integer.

tomkerkhove:
Doesn't sound ideal imo

stijnmoreels:
> Doesn't sound ideal imo

So... the current PR is good then for you?

tomkerkhove:
Not really, it should automatically add jobid for all messages AFAIK you can do that with ILogger

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -163,7 +192,7 @@ protected virtual async Task CompleteMessageAsync(string lockToken)
 
             if (_messageReceiver == null)
             {
-                throw new InvalidOperationException("Message receiver is not initialized yet.");
+                throw new InvalidOperationException($"[Job: {JobId}] Message receiver is not initialized yet");
```
tomkerkhove: We shouldn't add these prefixes to our exceptions

stijnmoreels: 👍

PR #80, opened 2020-03-11
Feature - make customized message pump still available with indroduction of message handlers

### LINKED ISSUE #43: Provide capability to control how JSON messages are deserialization
Provide capability to control how JSON messages are deserialization to match customer needs.

stijnmoreels:
We could also provide some custom return type if that's necessary or more user-friendly.
I think this (or something simulair) is required if we want to control JSON deserialization since in previous versions (without message handlers) we already could do that so we should keep that functionality.

Relates to #43

### DIFF
--- a/src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
+++ b/src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
@@ -4,13 +4,10 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Reflection;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Arcus.Messaging.Abstractions;
 using GuardNet;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
 namespace Arcus.Messaging.Pumps.Abstractions.MessageHandling
 {
@@ -26,10 +23,20 @@ namespace Arcus.Messaging.Pumps.Abstractions.MessageHandling
         {
             Guard.NotNull(serviceType, nameof(serviceType));
             Guard.NotNull(createMessageHandlerImplementation, nameof(createMessageHandlerImplementation));
+            Guard.For<ArgumentException>(
+                () => serviceType.GenericTypeArguments.Length != 2, 
+                $"Message handler type '{serviceType.Name}' has not the expected 2 generic type arguments");
 
             _serviceType = serviceType;
             _createMessageHandlerImplementation = createMessageHandlerImplementation;
+
+            MessageType = _serviceType.GenericTypeArguments[0];
         }
+
+        /// <summary>
+        /// Gets the type of the message that this abstracted message handler can process.
+        /// </summary>
+        internal Type MessageType { get; }
 
         /// <summary>
         /// Subtract all the <see cref="IMessageHandler{TMessage,TMessageContext}"/> implementations from the given <paramref name="serviceProvider"/>.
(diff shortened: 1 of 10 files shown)


stijnmoreels:
Like, if I want to deserialize CloudEvents.

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -26,11 +23,21 @@ private MessageHandler(Type serviceType, Func<object> createMessageHandlerImplem
         {
             Guard.NotNull(serviceType, nameof(serviceType));
             Guard.NotNull(createMessageHandlerImplementation, nameof(createMessageHandlerImplementation));
+            Guard.For<ArgumentException>(
```
tomkerkhove: This looks a bit strange, no? But must be reflection love 🤡

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -26,11 +23,21 @@ private MessageHandler(Type serviceType, Func<object> createMessageHandlerImplem
...
             _serviceType = serviceType;
             _createMessageHandlerImplementation = createMessageHandlerImplementation;
+
+            MessageType = _serviceType.GenericTypeArguments[0];
         }
 
+        /// <summary>
+        /// Gets the type of the message that this abstracted message handler can process.
+        /// </summary>
+        internal Type MessageType { get; }
```
tomkerkhove: If it's internal, do you need it as a property?

stijnmoreels: It's used in `MessagePump`. So we need access to it, but only internally.

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: It's an extra safeguard that only the open generic type is here added.

stijnmoreels:
Hmm, still a problem bc we can't add the message pump with the extensions and the settings to initialize the pump are internal.

stijnmoreels:
@tomkerkhove , I think there's still a problem with the connection string and how we use the `MessageReceiver` class instead of the `SubscriptionClient` for topic connection strings. The tests succeed, but the application isn't closing gracefully, leaving the topic subscription existing on the service bus resource after the test run.

src/Arcus.Messaging.Tests.Integration/Fixture/ServiceBusQueueAndTopicProgram.cs
```diff
@@ -45,7 +45,7 @@ public static void main(string[] args)
                     });
 
                     services.AddServiceBusQueueMessagePump(configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING_WITH_TOPIC"])
-                            .AddServiceBusTopicMessagePump("Receive-All", configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING_WITH_QUEUE"])
+                            .AddServiceBusTopicMessagePump("Test-Receive-All-Topic-And-Queue", configuration => configuration["ARCUS_SERVICEBUS_CONNECTIONSTRING_WITH_QUEUE"])
```
tomkerkhove: I presume it's on purpose that you pass queue connection info for topic?

stijnmoreels: Hmm, no 😄 .

stijnmoreels:
@tomkerkhove , I also added message handler overloads with a `Func<IServiceProvider, TMessagehandler>` so message handlers with extra parameters can take advantage of the `IServiceProvider` to retrieve additional services/logger/...

PR #84, opened 2020-03-21
Feature - use Observability correlation access

stijnmoreels:
Relates to https://github.com/arcus-azure/arcus.observability/pull/43

### DIFF
--- a/src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
+++ b/src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
@@ -31,12 +31,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Arcus.Observability.Correlation" Version="20200323.12.0-preview" />
+    <PackageReference Include="Arcus.Observability.Correlation" Version="0.1.0-preview-5" />
     <PackageReference Include="Guard.NET" Version="1.2.0" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
   </ItemGroup>
 
-   <PropertyGroup>
-    <RestoreSources>$(RestoreSources);https://api.nuget.org/v3/index.json;https://www.myget.org/F/arcus/api/v3/index.json</RestoreSources>
-  </PropertyGroup>
 </Project>
--- a/src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
+++ b/src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
@@ -36,4 +36,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
   </ItemGroup>
 
+   <PropertyGroup>
+    <RestoreSources>$(RestoreSources);https://api.nuget.org/v3/index.json;https://www.myget.org/F/arcus/api/v3/index.json</RestoreSources>
+  </PropertyGroup>
 </Project>
--- a/src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
+++ b/src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
@@ -31,7 +31,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Arcus.Observability.Correlation" Version="0.1.0-preview-2" />
+    <PackageReference Include="Arcus.Observability.Correlation" Version="20200323.12.0-preview" />
     <PackageReference Include="Guard.NET" Version="1.2.0" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
   </ItemGroup>
--- a/src/Arcus.Messaging.Pumps.Abstractions/Arcus.Messaging.Pumps.Abstractions.csproj
+++ b/src/Arcus.Messaging.Pumps.Abstractions/Arcus.Messaging.Pumps.Abstractions.csproj
@@ -39,4 +39,8 @@
     <ProjectReference Include="..\Arcus.Messaging.Abstractions\Arcus.Messaging.Abstractions.csproj" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <RestoreSources>$(RestoreSources);https://api.nuget.org/v3/index.json;https://www.myget.org/F/arcus/api/v3/index.json</RestoreSources>
+  </PropertyGroup>
+
 </Project>
--- a/src/Arcus.Messaging.Pumps.Abstractions/Arcus.Messaging.Pumps.Abstractions.csproj
+++ b/src/Arcus.Messaging.Pumps.Abstractions/Arcus.Messaging.Pumps.Abstractions.csproj
@@ -31,8 +31,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Arcus.Observability.Telemetry.Serilog.Enrichers" Version="20200323.12.0-preview" />
     <PackageReference Include="Guard.NET" Version="1.2.0" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="Serilog" Version="2.9.0" />
   </ItemGroup>
 
   <ItemGroup>
@@ -42,5 +44,5 @@
   <PropertyGroup>
     <RestoreSources>$(RestoreSources);https://api.nuget.org/v3/index.json;https://www.myget.org/F/arcus/api/v3/index.json</RestoreSources>
   </PropertyGroup>
-
+  
 </Project>
--- a/src/Arcus.Messaging.Pumps.Abstractions/Arcus.Messaging.Pumps.Abstractions.csproj
+++ b/src/Arcus.Messaging.Pumps.Abstractions/Arcus.Messaging.Pumps.Abstractions.csproj
@@ -31,7 +31,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Arcus.Observability.Telemetry.Serilog.Enrichers" Version="20200323.12.0-preview" />
+    <PackageReference Include="Arcus.Observability.Telemetry.Serilog.Enrichers" Version="0.1.0-preview-5" />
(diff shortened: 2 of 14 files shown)


stijnmoreels:
I'll add also an enricher for this.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -438,6 +441,11 @@ private async Task HandleMessageAsync(Message message, CancellationToken cancell
                 }
 
                 MessageCorrelationInfo correlationInfo = message.GetCorrelationInfo();
+                if (_correlationInfoAccessor != null)
+                {
+                    _correlationInfoAccessor.CorrelationInfo = correlationInfo;
```
tomkerkhove: The custom one will have to be able to coop with correlation per message instance, or we have to create a new instance of `ICorrelationInfoAccessor<MessageCorrelationInfo>` every time.

stijnmoreels: Yes yes, that's also something I mentioned at the very beginning.
I thought that having an extenion on the message itself was enough, but apperently not.
We could maybe not set it here like this, but wrap the `Message` in a inherited `CorrelatedMessage` or so?

stijnmoreels: If we would do that, we would indeed have no need for a 'Set' I think in the accessor.

tomkerkhove: I would not introduce a new message object as it will only be passed to the handlers and used by enrichers

stijnmoreels: Maybe some message id then to send to the correlation accessor and that it holds a concurrent dictionary or so.

tomkerkhove: That assumes the enricher knows the message id, which is not always the case

tomkerkhove: With enrichers you have to assume they have no knowledge of what is going on other than what is in the log entry

stijnmoreels: So... how would the enricher than now which message/correlation to write?

stijnmoreels: Because then the enricher can be called from another place, writing a entry about an other message, and assuming that the enricher can only call something without any identity, how can we be sure...
Maybe we should send something in the log message then, that the enricher can use to send to the enricher to retreive the correlation.

tomkerkhove: That's what we have to figure out! But I'd keep it very basic and straight forward; a POC outside of this PR should be done first imo.

stijnmoreels: Yet another branch? I think it would make it more complex.
We could have an issue for that, or I can work on this WIP PR further maybe.

I'll try to come up with the thing I said about the log message sending and ID.

tomkerkhove: Ok, fine by me.

I think in the end the best approach is a scoped dependency that is being used. Then it will be specific per message

stijnmoreels: Yes, but not accessable from the enricher, no? bc the scope will have to be created in the pump, and the enricher will not have reference to the service provider of the scope 😢 .

stijnmoreels: Made a proposal with the combination with scoped services and scoped logging.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -437,20 +440,29 @@ private async Task HandleMessageAsync(Message message, CancellationToken cancell
...
                 MessageCorrelationInfo correlationInfo = message.GetCorrelationInfo();
-                Logger.LogInformation(
-                    "Received message '{MessageId}' (Transaction: {TransactionId}, Operation: {OperationId}, Cycle: {CycleId})",
-                     message.MessageId, correlationInfo.TransactionId, correlationInfo.OperationId, correlationInfo.CycleId);
-                
-                var messageContext = new AzureServiceBusMessageContext(message.MessageId, message.SystemProperties,
-                    message.UserProperties);
+                using (IServiceScope serviceScope = ServiceProvider.CreateScope())
+                {
+                    var correlationInfoAccessor = serviceScope.ServiceProvider.GetService<ICorrelationInfoAccessor<MessageCorrelationInfo>>();
```
stijnmoreels: This would mean that the correlation accessor in observability should be registered as scoped.

src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
```diff
@@ -31,9 +31,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Arcus.Observability.Correlation" Version="0.1.0-preview-2" />
+    <PackageReference Include="Arcus.Observability.Correlation" Version="20200323.12.0-preview" />
```
tomkerkhove: Let's use the observation preview here

src/Arcus.Messaging.Pumps.Abstractions/Arcus.Messaging.Pumps.Abstractions.csproj
```diff
@@ -31,12 +31,18 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Arcus.Observability.Telemetry.Serilog.Enrichers" Version="20200323.12.0-preview" />
```
tomkerkhove: Same here

src/Arcus.Messaging.Pumps.ServiceBus/Arcus.Messaging.Pumps.ServiceBus.csproj
```diff
@@ -41,4 +41,8 @@
     <ProjectReference Include="..\Arcus.Messaging.ServiceBus.Core\Arcus.Messaging.ServiceBus.Core.csproj" />
   </ItemGroup>
 
+  <PropertyGroup>
```
tomkerkhove: Don't need this anymore I guess?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
tomkerkhove: That's the case now so resolving, ok?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -437,20 +440,29 @@ private async Task HandleMessageAsync(Message message, CancellationToken cancell
...
+                    correlationInfoAccessor.SetCorrelationInfo(correlationInfo);
+
+                    using (LogContext.Push(new MessageCorrelationInfoEnricher(correlationInfoAccessor)))
+                    {
+                        Logger.LogInformation(
+                            "Received message '{MessageId}' (Transaction: {TransactionId}, Operation: {OperationId}, Cycle: {CycleId})",
```
tomkerkhove: You should no longer need this:
```suggestion
                            "Received message '{MessageId}'",
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -437,20 +440,29 @@ private async Task HandleMessageAsync(Message message, CancellationToken cancell
...
+                            message.MessageId, correlationInfo.TransactionId, correlationInfo.OperationId, correlationInfo.CycleId);
```
tomkerkhove: You should no longer need this:
```suggestion
                            message.MessageId);
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -437,20 +440,29 @@ private async Task HandleMessageAsync(Message message, CancellationToken cancell
...
-                     message.MessageId, correlationInfo.TransactionId, correlationInfo.OperationId, correlationInfo.CycleId);
-                
-                var messageContext = new AzureServiceBusMessageContext(message.MessageId, message.SystemProperties,
-                    message.UserProperties);
+                using (IServiceScope serviceScope = ServiceProvider.CreateScope())
+                {
+                    var correlationInfoAccessor = serviceScope.ServiceProvider.GetService<ICorrelationInfoAccessor<MessageCorrelationInfo>>();
+                    correlationInfoAccessor.SetCorrelationInfo(correlationInfo);
+
+                    using (LogContext.Push(new MessageCorrelationInfoEnricher(correlationInfoAccessor)))
```
tomkerkhove: Just curious - Why do we need to use this approach?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Yes indeed

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: It makes enrichment of logging also just like the correlation scoped so we can benefit from the different kinds of correlation on message level.

Issue #26, opened 2019-12-04
Provide capability to support multiple message handlers based on message context


tomkerkhove:
Provide capability to support multiple message handlers based on message context. This would allow users to map inbound messages to different message handlers in the same compute unit.

This is important if different message types are shared on the same queue, migrating to a new message contract, etc.

It's up to the users to map the message to the message handler that should be called, this can be as simple as the following:
```csharp
.AddServiceBusPump()
   .WithMessagHandler<OrderV1MessageHandler>(messageContext => messageContext.MessageType == MessageTypes.OrderV1);
   .WithMessagHandler<OrderV2MessageHandler>(messageContext => messageContext.MessageType == MessageTypes.OrderV2);
```

But this is just pseudo code to prove the point, can also be an action that is being passed when creating the pump.

PR #91, opened 2020-06-05
Feature - add message context filter to the message handler registration

stijnmoreels:
Relates to #26


stijnmoreels:
Seems like the build is failing on master too, so could be that it's not related with these changes here.

docs/preview/features/message-pumps/customization.md
```diff
@@ -40,4 +40,34 @@ public class OrderMessagePump : MessagePump
...
+We would like that this handler only processed the message when the context contains `MessageType` equals `Order`.
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.WithMessageHandler<OrderMessageHandler, Order>(context => context.Properties["MessageType"].ToString() == "Order");
+}
+```
+
+> Note that the order in which the message handlers are registered is importent in the message processing.
```
tomkerkhove, line 70:
```suggestion
> Note that the order in which the message handlers are registered is important in the message processing.
```

docs/preview/features/message-pumps/customization.md
```diff
@@ -40,4 +40,34 @@ public class OrderMessagePump : MessagePump
 }
 ```
 
+## Control message context filtering
+
+When registering a new `IMessageHandler<>`, one can choose to add an additional filter on the message context used in that handler.
```
tomkerkhove, line 45: What about this?
```suggestion
When registering a new message handler, one can opt-in to add a filter on the message context which filters out messages that are not needed to be processed.

This can be useful when you are sending different message types on the same queue. Another use-case is being able to handle different versions of the same message type which have different contracts because you are migrating your application.
```

docs/preview/features/message-pumps/customization.md
```diff
@@ -40,4 +40,34 @@ public class OrderMessagePump : MessagePump
 }
 ```
 
+## Control message context filtering
```
tomkerkhove, line 43: What do you think of one of these?
```suggestion
## Filter messages based on message context
```

```suggestion
## Process subset of messages with message context filtering
```

docs/preview/features/message-pumps/customization.md
```diff
@@ -40,4 +40,34 @@ public class OrderMessagePump : MessagePump
...
+When registering a new `IMessageHandler<>`, one can choose to add an additional filter on the message context used in that handler.
+
+Following example shows how a message handler should only process a certain message when a property's in the context is present.
+
+We'll use a simple message handler implementation:
+
+```csharp
+public class OrderMessageHandler : IMessageHandler<Order>
+{
+    public async Task ProcessMessageAsync(Order order, MessageContext context, ...)
```
tomkerkhove, line 54: It's been a while so here is a stupid question - They were able to use different MessageContext type as well, right? Like the Service bus one?

stijnmoreels: Not a stupid question 😉 , yest that's true, one can register message handlers with custom and built-in message context, altough I didn't implement or test this just yet in this PR. I thought it was a good start to first begin with the 'baseline' for this message context filtering.

Also why this PR is not closing the related issue 🤓 .

stijnmoreels:
> This looks good to me, added a few comments.
> 
> Can we introduce some integration tests for this please?

Yes, of course. What would you have in mind?
* Registering two handlers, only one with the correct context type
* Registering multiple handlers with the same context type, only one with the correct context predicate

stijnmoreels:
> > This looks good to me, added a few comments.
> > Can we introduce some integration tests for this please?
> 
> Yes, of course. What would you have in mind?
> 
> * Registering two handlers, only one with the correct context type
> * Registering multiple handlers with the same context type, only one with the correct context predicate

Ok, I added these two tests + required me to add the extensions on the service bus message handlers, so that's added no too.

docs/preview/features/message-pumps/customization.md
tomkerkhove: Ok, thanks!

tomkerkhove:
> > This looks good to me, added a few comments.
> > Can we introduce some integration tests for this please?
> 
> Yes, of course. What would you have in mind?
> 
> * Registering two handlers, only one with the correct context type
> * Registering multiple handlers with the same context type, only one with the correct context predicate

Just some tests to verify that the correct handler is processing messages if the criteria is met

stijnmoreels:
> > > This looks good to me, added a few comments.
> > > Can we introduce some integration tests for this please?
> > 
> > 
> > Yes, of course. What would you have in mind?
> > 
> > * Registering two handlers, only one with the correct context type
> > * Registering multiple handlers with the same context type, only one with the correct context predicate
> 
> Just some tests to verify that the correct handler is processing messages if the criteria is met

Yes, added two tests to test selection based on type and predicate.

docs/preview/features/message-pumps/customization.md
stijnmoreels: But as explained below, I added some extensions for this too bc I needed it for the integration tests.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -728,5 +728,49 @@ public static class IServiceCollectionExtensions
...
+
+        /// <summary>
+        /// Adds a <see cref="IAzureServiceBusMessageHandler{TMessage}" /> implementation to process the messages from an <see cref="AzureServiceBusMessagePump"/> implementation.
+        /// resources.
+        /// </summary>
+        /// <typeparam name="TMessageHandler">The type of the implementation.</typeparam>
+        /// <typeparam name="TMessage">The type of the message that the message handler will process.</typeparam>
+        /// <param name="services">The collection of services to use in the application.</param>
+        /// <param name="messageContextFilter">The function that determines if the message handler should handle the message based on the context.</param>
+        public static IServiceCollection WithServiceBusMessageHandler<TMessageHandler, TMessage>(
```
tomkerkhove, line 740: We should decouple this from Service Bus and only give them `MessageContext`

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -728,5 +728,49 @@ public static class IServiceCollectionExtensions
...
+        /// <summary>
+        /// Adds a <see cref="IAzureServiceBusMessageHandler{TMessage}" /> implementation to process the messages from an <see cref="AzureServiceBusMessagePump"/> implementation.
+        /// resources.
+        /// </summary>
+        /// <typeparam name="TMessageHandler">The type of the implementation.</typeparam>
+        /// <typeparam name="TMessage">The type of the message that the message handler will process.</typeparam>
+        /// <param name="services">The collection of services to use in the application.</param>
+        /// <param name="messageContextFilter">The function that determines if the message handler should handle the message based on the context.</param>
+        /// <param name="implementationFactory">The function that creates the message handler.</param>
+        public static IServiceCollection WithServiceBusMessageHandler<TMessageHandler, TMessage>(
```
tomkerkhove, line 762: We should decouple this from Service Bus and only give them `MessageContext`

stijnmoreels: Why is that? Isn't that the point of registering a service bus message handler? If people want a different context, shouldn't they use the `.WithMessageHandler` instead?

tomkerkhove: This is what we want to avoid:
image

Meaning, people need to have the opportunity to be able to use message handler with filtering that does not tightly-couple it to Service Bus. Do that mean we should remove the current on? No, but we need to have the filtering capability on `WithMessageHandler` at least.

That's why we decoupled it a while back, to make sure the same message handler can be used with another queuing tech.

stijnmoreels: Yes, and the context filtering is available on the `.WithMessageHandler`, the `.WithServiceBusMessageHandler` only adds the specific service bus message context to the handler so we don't have to specify:

```csharp
.WithMessageHandler<OrdersMessageHandler, Order, ServiceBusMessageContext>()
```

stijnmoreels: So, I'm not sure what I should remove/change here and if I implement something wrong here maybe?

tomkerkhove: I've misread, didn't notice the other extensions sorry! No action to be taken then but I would at least add some tests using the non-SB approach since we're replacing those

stijnmoreels: Yes, updated the existing test to verify more thoroughly the predication selection.

stijnmoreels:
> I might be taking this a bit too far but I'm missing an integration test where we:
> 
> 1. Send a message with a given property in context
> 2. Have multiple handlers configured emitting different events
> 3. We ensure we receive the expected event type from the expected handler filtering out those messages
> 
> Or did I miss it?

This is kinda done with my recent update of the predicate selection tests where 3 message handlers are configured, (1 being without the specific Service Bus extension) and 2 with different message types and predicate selection on a property on the message context.

PR #107, opened 2020-08-24
Feature - rotate message pump on SecretNewVersionCreated event

### LINKED ISSUE #101: Provide capability to restart message pump if secret for connection string was updated
Provide the capability to restart message pump if secret for connection string was updated so that we don't have to wait for unauthorized exceptions and we always use the connection string that should be used.

When this is the case, new messages should no longer be received and currently being processed messages should be completed instead of abrupt shutdown.

Based on Key Vault events we see that the secret containing the connection string was changed after which the host restarts itself.

Wait for unauthorized exceptions, but that doesn't cover the scenario where the app is assigned to a new service bus namespace or less permissions.

**Additional context**
I'm not sure if this should be the default or through opt-in since this can lead to two restarts when using sec/primary rotation and assignments.

stijnmoreels:
Because the `Arcus.Messaging.Pumps.ServiceBus` project was used in the `Arcus.BackgroundJobs.CloudEvents`, we couldn't reference the CloudEvents pacakge inside the Azure Service Bus project due to circular references.
That's why a new project is created called `Arcus.Messaging.Pumps.ServiceBus.KeyRotation` (please help me with the name of this thing 😅) while this problem is resolved.

We could create some CloudEvent message pump implementation ourselves so we can include this inside the same project, but maybe we're missing an opportiunity here.
By placing it in another project, exposing the necessary operations on the `AzureServiceBusMessagePump` to restart the pump, we could allow other devs to create additional 'restart message handlers' if they want. (For example: when an secret is updated in HashiCorp, and you want to restart the pump).

Now, we have an additional message pump that will listen for `SecretNewVersionCreated` events and will process this inside the `ReauthenticateMessageHandler` which will restart the message pump. How this message handlers knowns _which_ message pump to restart is solved by the `JobId` which identifies the message pump.

An additional `MaxUnauthorizedExceptionsCount` is added to the options so devs can choose how long the message pump must ignore unauthorized exceptions (bc waiting for the `SecretNewVersionCreated` event), before restarting the message pump itself. This is actually a fallback so we benefit from both systems.

Closes #101


stijnmoreels:
Hold off with docs bc I figured that this is not something you'll fully agree on 😅 .

stijnmoreels:
Maybe even moved to background jobs.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Arcus.Messaging.Pumps.ServiceBus.KeyRotation.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageId>Arcus.Messaging.Pumps.ServiceBus.KeyRotation</PackageId>
+    <Authors>Arcus</Authors>
+    <Company>Arcus</Company>
+    <Product>Arcus.Messaging</Product>
+    <Description>Provides capability related to key rotation of Azure Service Bus message pumps</Description>
```
tomkerkhove, line 9:
```suggestion
    <Description>Provides capability related to the resiliency of Azure Service Bus message pumps in case of key rotation</Description>
```

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+            services.AddCloudEventBackgroundJob(subscriptionNamePrefix, serviceBusTopicConnectionStringSecretKey);
+            services.WithServiceBusMessageHandler<ReAuthenticateMessageHandler, CloudEvent>(serviceProvider =>
+            {
+                AzureServiceBusMessagePump messagePump =
+                    serviceProvider.GetServices<IHostedService>()
+                                   .OfType<AzureServiceBusMessagePump>()
+                                   .FirstOrDefault(pump => pump.JobId == jobId);
+
+                Guard.NotNull(messagePump, nameof(messagePump), 
+                    $"Cannot register re-authentication without a '{nameof(AzureServiceBusMessagePump)}' with 'JobId' = '{jobId}'");
```
tomkerkhove, line 56:
```suggestion
                    $"Cannot register re-authentication without an {nameof(AzureServiceBusMessagePump)} with job id {{{jobId}}}", jobId);
```

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+
+                messagePump.Settings.Options.MaximumUnauthorizedExceptionsBeforeRestart = maximumUnauthorizedExceptionsBeforeRestart;
```
tomkerkhove, line 58: Should we mutate the message pump here?

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateMessageHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+            var secretNewVersionCreated = message.GetPayload<SecretNewVersionCreated>();
+            if (secretNewVersionCreated is null)
+            {
+                throw new CloudException(
+                    "Azure Key Vault job cannot map Event Grid event to CloudEvent because the event data isn't recognized as a 'SecretNewVersionCreated' schema");
+            }
+
+            _logger.LogInformation("Received Azure Key vault 'Secret New Version Created' event");
+            
+            await _messagePump.RestartAsync();
```
tomkerkhove, line 69: Hold your horses 😁 You only need to restart it when the secret name matches the one that is being used to authenticate the message pump with 😅

src/Arcus.Messaging.Tests.Integration/Fixture/KeyRotationConfig.cs
```diff
@@ -10,24 +11,27 @@ public class KeyRotationConfig
...
+            KeyVault = keyVault;
             ServicePrincipal = servicePrincipal;
             ServiceBusNamespace = serviceBusNamespace;
         }
 
         /// <summary>
         /// Gets the config representing a Azure Key Vault secret.
         /// </summary>
-        public KeyVaultSecret KeyVaultSecret { get; }
+        public KeyVault KeyVault { get; }
```
tomkerkhove, line 34: Isn't this more clear?
```suggestion
        public KeyVaultConfig KeyVault { get; }
```

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateMessageHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Rest.Azure;
+
+namespace Arcus.Messaging.Pumps.ServiceBus.KeyRotation
+{
+    /// <summary>
+    /// Represents an <see cref="IMessageHandler{TMessage, TMessageContext}"/> that processes <see cref="CloudEvent"/>s representing <see cref="SecretNewVersionCreated"/> events
+    /// that will eventually result in restarting an <see cref="AzureServiceBusMessagePump"/> instance.
+    /// </summary>
+    public class ReAuthenticateMessageHandler : IAzureServiceBusMessageHandler<CloudEvent>
```
tomkerkhove, line 18: It's a long name, but maybe more explicit?
```suggestion
    public class MessagePumpAuthenticationRotatedMonitorMessageHandler : IAzureServiceBusMessageHandler<CloudEvent>
```

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="maximumUnauthorizedExceptionsBeforeRestart">
+        ///     The fallback when the Azure Key Vault notification doesn't get delivered correctly,
+        ///     how many times should the message pump run into an <see cref="UnauthorizedException"/> before restarting.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        ///     Thrown when the <paramref name="services"/> or the searched for <see cref="AzureServiceBusMessagePump"/> based on the given <paramref name="jobId"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="subscriptionNamePrefix"/> or <paramref name="serviceBusTopicConnectionStringSecretKey"/> is blank.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="maximumUnauthorizedExceptionsBeforeRestart"/> is less then zero.</exception>
+        public static IServiceCollection WithReAuthenticationOnNewSecretVersion(
```
tomkerkhove, line 35: What if it's called this? Maybe feels more natural?
```suggestion
        public static IServiceCollection WithAutoRestartOnRotatedCredentials(
```

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
tomkerkhove: If we do, what if we call a "reconfigure" method instead of changing the options directly?

tomkerkhove (commented):
It's definately a good start! Feel free to let me know if the docs are there to have an end-user perspective how to use it.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateMessageHandler.cs
stijnmoreels: 😄 that's the burden of OOP , long names 😉 . We could rename it but let me think of something 🤔 .

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Ah, yes, that's good!

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: This is not a log event, this is a guard. Can't do that here.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: OK, yes, done that.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateMessageHandler.cs
stijnmoreels: Renamed to `ReAuthenticateOnRotatedCredentialsMessageHandler`.

stijnmoreels:
> It's definately a good start! Feel free to let me know if the docs are there to have an end-user perspective how to use it.

OK, added some feature docs to explain how this works and how to set up.

docs/preview/features/message-pumps/service-bus.md
```diff
@@ -302,6 +303,58 @@ string transactionId = correlationInfo.TransactionId;
...
 
+## Automatic Azure Key Vault credentials rotation
+
+The additional library `Arcus.Messaging.Pumps.ServiceBus.KeyRotation` provides an extension on the message pump to restart the pump automatically when the credentials of the pump stored in Azure Key Vault are changed.
+This feature allows more reliable restarting instead of relying on authentication exceptions that may be throwed during the lifetime of the message pump.
+
+## How does this work?
+
+A background job will be run in the background which is in fact a message pump on a Service Bus Topic.
+The Azure Key Vault where the connection string for your target message pump is located should send `SecretNewVersionCreated` events to this Service Bus Topic.
```
tomkerkhove, line 314: What about this?
```suggestion
A background job is polling for `SecretNewVersionCreated` events on an Azure Service Bus Topic for the secret that stores the connection string.
```

docs/preview/features/message-pumps/service-bus.md
```diff
@@ -302,6 +303,58 @@ string transactionId = correlationInfo.TransactionId;
...
+
+That way, when the background job receives a new Key Vault event, it will know that your target message pump should be restarted.
```
tomkerkhove, line 316:
```suggestion
That way, when the background job receives a new Key Vault event, it will get the latest connection string, restart the message pump and authenticate with the latest credentials.
```

docs/preview/features/message-pumps/service-bus.md
tomkerkhove: The reason is that the rest of it is an implementation detail that people don't care about, imo

docs/preview/features/message-pumps/service-bus.md
```diff
@@ -302,6 +303,58 @@ string transactionId = correlationInfo.TransactionId;
...
+{
+    // You should have a unique Job ID to identity the message pump so the automatic process knows which pump to restart.
+    string jobId = Guid.NewGuid().ToString();
+
+    string secretName = hostContext.Configuration["ARCUS_KEYVAULT_CONNECTIONSTRINGSECRETNAME"];
+    services.AddServiceBusQueueMessagePump(secretName, options => options.JobId = jobId)
+            .WithServiceBusMessageHandler<OrdersAzureServiceBusMessageHandler, Order>();
+
+    // This extension will be available to you once you installed the package.
+    services.WithAutoRestartOnRotatedCredentials(
```
tomkerkhove, line 341: I would use this:
```suggestion
    services.WithAutoRestartServiceBusQueueMessagePumpOnRotatedCredentials(
```

Otherwise, it's not clear what you are autorestarting since it's not linked to https://github.com/arcus-azure/arcus.messaging/pull/107/files#diff-1f44b85c1f99fedb43b7e3a702a84b81d8ca6b38533d1a607ecec94295e9cc09R337

tomkerkhove: We'll reduce it again as part of https://github.com/arcus-azure/arcus.messaging/issues/152

docs/preview/features/message-pumps/service-bus.md
```diff
@@ -302,6 +303,58 @@ string transactionId = correlationInfo.TransactionId;
...
+        jobId: jobId, 
+        subscriptionNamePrefix: "TestSub", 
+        
+        // The secret key where the Azure Service Bus Topic connection string is located that the background job will use to receive the Azure Key vault events.
+        serviceBusTopicConnectionStringSecretKey: "ARCUS_KEYVAULT_SECRETNEWVERSIONCREATED_CONNECTIONSTRING",
+        
+        // The secret key where the Azure Service Bus connection string is located that your target message pump uses.
+        // This secret key name will be used to check if the received Azure Key Vault event is from this secret or not.
+        targetConnectionStringKey: secretName,
```
tomkerkhove, line 350:
```suggestion
        messagePumpConnectionStringKey: secretName,
```

But ideally we pull this in automatically since we know it already as part of https://github.com/arcus-azure/arcus.messaging/pull/107/files#diff-1f44b85c1f99fedb43b7e3a702a84b81d8ca6b38533d1a607ecec94295e9cc09R337

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -110,20 +118,28 @@ public static IEnumerable<MessageHandler> SubtractFrom(IServiceProvider serviceP
...
 
                     var canProcessMessage = (bool) Service.InvokeMethod(
                         "CanProcessMessage",
                         BindingFlags.Instance | BindingFlags.NonPublic,
                         messageContext);
 
-                    _logger.LogInformation(
-                        "Message context predicate registered with the message handler {MessageHandlerType} resulted in {Result}, so {Action} process this message",
-                        ServiceType.Name, canProcessMessage, canProcessMessage ? "can" : "can't");
+                        _logger.LogInformation(
```
tomkerkhove, line 132:
```suggestion
                        _logger.LogTrace(
```

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            string serviceBusTopicConnectionStringSecretKey,
+            string targetConnectionStringKey)
+        {
+            Guard.NotNull(services, nameof(services), "Requires a collection of services to add the re-authentication background job");
+            Guard.NotNullOrWhitespace(subscriptionNamePrefix, nameof(subscriptionNamePrefix), "Requires a non-blank subscription name of the Azure Service Bus Topic subscription, to receive Azure Key Vault events");
+            Guard.NotNullOrWhitespace(serviceBusTopicConnectionStringSecretKey, nameof(serviceBusTopicConnectionStringSecretKey), "Requires a non-blank secret key that points to a Azure Service Bus Topic");
+            Guard.NotNullOrWhitespace(targetConnectionStringKey, nameof(targetConnectionStringKey), "Requires a non-blank secret key that points to the credentials that holds the connection string of the target message pump");
+
+            services.AddCloudEventBackgroundJob(subscriptionNamePrefix, serviceBusTopicConnectionStringSecretKey);
+            services.WithServiceBusMessageHandler<ReAuthenticateOnRotatedCredentialsMessageHandler, CloudEvent>(serviceProvider =>
```
tomkerkhove, line 44: Are we sure this doesn't impact our other message pump?

This might be another argument for https://github.com/arcus-azure/arcus.messaging/issues/152 to ensure that.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateOnRotatedCredentialsMessageHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+            CancellationToken cancellationToken)
+        {
+            Guard.NotNull(message, nameof(message), "Cannot invalidate Azure KeyVault secret from a 'null' CloudEvent");
+
+            _logger.LogTrace("Receiving new Azure Key Vault notification...");
+            
+            var secretNewVersionCreated = message.GetPayload<SecretNewVersionCreated>();
+            if (secretNewVersionCreated is null)
+            {
+                throw new CloudException(
```
tomkerkhove, line 67: Throw exception or just ignore? Because if you are using the same topic for other events you'll definitely be throwing this exception, without a reason?

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateOnRotatedCredentialsMessageHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+                    "Azure Key Vault job cannot map Event Grid event to CloudEvent because the event data isn't recognized as a 'SecretNewVersionCreated' schema");
+            }
+
+            if (_targetConnectionStringKey == secretNewVersionCreated.ObjectName)
+            {
+                _logger.LogInformation("Received Azure Key vault 'Secret New Version Created' event, restarting target message pump {JobId}", _messagePump.JobId);
```
tomkerkhove, line 73: This is maybe a verbose one which I would combine with an event, personally

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateOnRotatedCredentialsMessageHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+                await _messagePump.RestartAsync();
+            }
+            else
+            {
+                _logger.LogTrace("Received Azure Key Vault 'Secret New Version Created' event but for wrong message pump");
```
tomkerkhove, line 78:
```suggestion
                _logger.LogTrace("Received Azure Key Vault 'Secret New Version Created' event for another secret, ignoring.");
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -309,17 +398,17 @@ private MessageHandlerOptions DetermineMessageHandlerOptions(AzureServiceBusMess
...
-                await CloseMessageReceiverAsync();
-                await OpenNewMessageReceiverAsync();
-                Logger.LogInformation("Azure Service Bus restarted!");
-            }
+            Interlocked.Exchange(ref _unauthorizedExceptionCount, 0);
+
+            Logger.LogTrace("Restarting Azure Service Bus...");
+            await CloseMessageReceiverAsync();
+            await OpenNewMessageReceiverAsync();
+            Logger.LogInformation("Azure Service Bus restarted!");
```
tomkerkhove, line 411: I would add message pump here, and optionally more information such as the name of entity being processed, job, etc.
```suggestion
            Logger.LogInformation("Azure Service Bus message pump restarted!");
```

That would help separating pumps that are constantly restarting compared to others

src/Arcus.Messaging.Tests.Integration/Fixture/KeyVaultConfig.cs
```diff
@@ -30,5 +36,10 @@ public KeyVaultSecret(string vaultUri, string secretName)
         /// Gets the name of the secret in the Azure Key Vault instance.
         /// </summary>
         public string SecretName { get; }
+
+        /// <summary>
+        /// Gets the event endpoint of the Azure Key Vault 'Secret new version created' event.
```
tomkerkhove, line 41:
```suggestion
        /// Gets the endpoint where Azure Key Vault events will be available, including 'Secret new version created' event.
```

tomkerkhove (changes requested):
Very good start, but needs a few tweaks imo

docs/preview/features/message-pumps/service-bus.md
stijnmoreels: If we'll reduce it, should we change it now?

docs/preview/features/message-pumps/service-bus.md
stijnmoreels: Hmm, I think the link isn't pointing to what you want it to point.

tomkerkhove: I should have used a commit but basically you know the secret already because it was passed to the pump

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: When other background jobs or message pumps are receiving cloud events, yes, this message handler will be one of the registered ones, but 1) it filters on secret new version created events (and with the message body filter we can even stricken it further), and 2) it uses the ID of the message pump to make sure that only the correct pump is restarted.
There exists a problem though, when multiple pumps/jobs are receiving cloud events and want to auto restart. Then, the consumer should stricken the cloud event filter, I guess.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateOnRotatedCredentialsMessageHandler.cs
stijnmoreels: Not sure what you mean 😅 .

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Or, we should have a way to add the job ID to the messaging context, but that would mean we should have access to how the cloud events background job create messaging contexts.

docs/preview/features/message-pumps/service-bus.md
stijnmoreels: That's not entirely correct, as it's used directly inline with `await secretProvider.GetRawSecretAsync(secretName)` the secret name is not saved (yet) in the settings. We could do that off course.

stijnmoreels: And also, when the user used the `Func<ISecretProvider, Task<string>>` we don't know what the secret is...

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateOnRotatedCredentialsMessageHandler.cs
stijnmoreels: Removed this. Resolving.

stijnmoreels:
Oh yeah, the build is green again. Means that the Azure Key Vault events are coming through again.

docs/preview/features/message-pumps/service-bus.md
stijnmoreels: Oh, I see now, we're gonna make this one obsolete when we introduce a new one for the fluent type. Got it.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: There is not entity being processed at this point, this is called upon a receival of an event or when the we receive unauthorized failures in the message pump.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: But yes, when we do #152 we can make sure that we have a separate `IServiceCollection` for each message router.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateOnRotatedCredentialsMessageHandler.cs
tomkerkhove: I would change this to Trace, but maybe call LogEvent afterwards as well ~ "Message Pump Restarted" or so?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
tomkerkhove: Can you elaborate on that please? I don't think I get what you're saying.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -309,17 +398,17 @@ private MessageHandlerOptions DetermineMessageHandlerOptions(AzureServiceBusMess
             return messageHandlerOptions;
         }
 
-        private async Task ReAuthenticateMessageReceiverAsync(Exception exception)
+        /// <summary>
+        /// Restart core functionality of the message pump.
+        /// </summary>
+        public async Task RestartAsync()
         {
-            if (exception is UnauthorizedException)
```
tomkerkhove, line 314: I don't know if we should remove this check maybe? While this is a great feature, we should not enforce people to use it, no?

src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePumpTests.cs
```diff
@@ -280,11 +281,53 @@ await using (var service = await TestMessagePumpService.StartNewAsync(config, _l
...
+                CommandArgument.CreateSecret("ARCUS_KEYVAULT_VAULTURI", keyRotationConfig.KeyVault.VaultUri),
+                CommandArgument.CreateSecret("ARCUS_KEYVAULT_CONNECTIONSTRINGSECRETNAME", keyRotationConfig.KeyVault.SecretName),
+                CommandArgument.CreateSecret("ARCUS_KEYVAULT_SERVICEPRINCIPAL_CLIENTID", keyRotationConfig.ServicePrincipal.ClientId),
+                CommandArgument.CreateSecret("ARCUS_KEYVAULT_SERVICEPRINCIPAL_CLIENTSECRET", keyRotationConfig.ServicePrincipal.ClientSecret),
+                CommandArgument.CreateSecret("ARCUS_KEYVAULT_SECRETNEWVERSIONCREATED_CONNECTIONSTRING", keyRotationConfig.KeyVault.SecretNewVersionCreated.ConnectionString)
+            };
+
+            using (var project = await WorkerProject.StartNewWithAsync<ServiceBusQueueSecretNewVersionReAuthenticateProgram>(config, _logger, commandArguments))
+            {
+                string newSecondaryConnectionString = await client.RotateConnectionStringKeysForQueueAsync(KeyType.SecondaryKey);
```
tomkerkhove, line 312: What if we run await service.SimulateMessageProcessingAsync before rotating it? That makes sure it is first using the old connection string and works fine.

After that, we rotate the connection and run it again which should be fine as well.

tomkerkhove (commented):
Almost there, added a few comments

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/Extensions/IServiceCollectionExtensions.cs
tomkerkhove: Let's keep it as is and #152 will fix it.

docs/preview/features/message-pumps/service-bus.md
tomkerkhove: Ok, fair enough.

src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePumpTests.cs
stijnmoreels: Hmm, isn't that what the other tests are for? It's some while ago when I wrote this, but I think I chose to rotate first because that would make sure that I got a connection string that's still valid.
I guess this test is more about restarting and the final check about simulating a message is the assertion-proof that after all the key rotations and restarts we can still receive messages.

We rotate 3 times here in the test. Something we agreed on was a concrete example of how an application would rotate it.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Aha, hmm, maybe make this a nullable integer so we only use the feature if the user set this value?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: This method where your comment is marked (`RestartAsync`) is called in two places:
1. When we receive unauthorized failures from the `MessageReceiver` itself, or
2. When we auto-restart upon the receival of `SecretNewVersionCreated` Key Vault events.

Both places don't interact with received messages. Ah, but now that I re-read your comment, you probably mean the `EntityPath` property, I thought you meant the message itself 😅 .
I'll change that!

stijnmoreels: Resolving bc I added these extra information like the other tracing.

src/Arcus.Messaging.Pumps.ServiceBus.KeyRotation/ReAuthenticateOnRotatedCredentialsMessageHandler.cs
stijnmoreels: Yes, did that! Resolving!

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
tomkerkhove: Not sure if I'm following because this is related to UnauthorizedException?

stijnmoreels: This is changed to a system where we count the amount of `UnauthorizedException`'s that's being thrown. A new configuration option is added that the user can configure. Before this PR we always restarted when we received a new `UnauthorizedException` but now we have to wait till the "Secret New Version Created" event is raised (when configured); so the user should also indicate how long they should wait for such event. That's why a new configuration option is added (`integer`). (also explained in the description of this PR).

My question here was, maybe we still check for `UnauthorizedException` types and count the amount of exceptions being thrown, but only when there's such a value present. AKA making it nullable.

stijnmoreels: Although, I think I'm not such a fan of removing the feature like it is now (by default), since this way we can use both features at the same time by default (default 5 `UnauthorizedException`'s can be throwed before the pump restarts).

tomkerkhove: Well for me it's important that we don't force people to rely on Azure Event Grid, but just give them the option. So I would support both

stijnmoreels: Not sure how Azure Event Grid comes into play here, as these Key Vault events are put on a Service Bus Topic. To get back to your original question, if we support them both, then we should indeed leave it the way it is. But can maybe be discussed offline.

tomkerkhove: Yes but Event Grid puts them on the topic, but that was indeed what I was referring to.

tomkerkhove: Feel free to resolve if both are supported now

tomkerkhove (approved):

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: The key vault event is optional, and can be combined with the already restart functionality, as explained in the meeting we had. I'll resolve this. You can look it over in mor detail once you have the time and we can do another checkup if needed.

PR #108, opened 2020-08-24
Publish docs for v0.2.0

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>


docs/features/message-pumps/service-bus.md
```diff
@@ -147,6 +151,10 @@ public void ConfigureServices(IServiceCollection services)
             // The unique identifier for this background job to distinguish 
             // this job instance in a multi-instance deployment (default: guid).
             options.JobId = Guid.NewGuid().ToString();
+
+            // The timeout when the message pump tries to restart and re-authenticate during key rotation of the connection string.
+            // (default: 5 seconds)
+            options.KeyRotationTimeout = TimeSpan.FromSeconds(10);
```
stijnmoreels, line 157: Woops, apperently we've forgot to remove this one from the setings and the docs bc it is now not used anymore.
Hmm, what whould be the best approach here?

docs/v0.2.0/index.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Support for using message pumps for the following brokers:
+    - Azure Service Bus ([docs](features/message-pumps/service-bus))
+    - Azure Service Bus Extensions ([docs](features/service-bus))
+- Support for exposing TCP health probes to periodically check liveness/readiness of the host ([docs](features/tcp-health-probe))
+- Customize message pumps ([docs](features/message-pumps/customization))
+
+## Older versions
+
+- [v0.1.0](v0.1.0)
```
stijnmoreels, line 26: This path has to be a path higher instead, otherwise one will end up with `/v0.2.0/v0.1.0`.

docs/features/message-pumps/service-bus.md
stijnmoreels: We could maybe still use this timeout in another way and pass along a cancellation token but not sure if that's added value with the current setup.

stijnmoreels (changes requested):
Path relocation and key rotation timeout that isn't being used anymore due to a change in implementation. We could still use it and pass along a cancellation token in the asynchronous methods but not sure if that's added value with the current setup.

docs/features/message-pumps/service-bus.md
tomkerkhove: No worries, I just copied the preview folder but will remove it. What's the right docs for this or was it simply removed?

docs/v0.2.0/index.md
tomkerkhove: Good catch, thanks!

docs/features/message-pumps/service-bus.md
stijnmoreels: It should have been removed, but it wasn't, so I guess it's also shipped, but doesn't do anything. 😓 .
Hmm, we could remove it from the docs and make it obsolete or something?

tomkerkhove: Fine by me!

tomkerkhove: We can still look at supporting it in a future release as well of course. What's the reason that it's not used?

stijnmoreels: Bc we first used a different approach in restarting the message pump, which the cancellation token played a part in, but with this new appraoch, we actually restart the internal 'kitchen' of the message pump, and then the cancellation token is not needed.

stijnmoreels: So, for this PR, I'll say it's okay?

stijnmoreels (approved):

docs/features/message-pumps/service-bus.md
tomkerkhove: If it's not used I'd remove it as I already did. If the docs are ok for you, feel free to merge!

PR #111, opened 2020-08-26
Feature - add fallback message handler as safety net during message handing

### LINKED ISSUE #110: Provide catch-all message handler
Provide a catch-all message handler that can be used in case no matching message handler was found.

stijnmoreels:
Adds an extra interface `IFallbackMessageHandler` that will handle remaining messages if none of the message handlers were able to handle the message.
- Added service collection extension to register this more easily
- Added section in customization docs to explain this feature
- Added integration test that will explicitly test the fallback of the message to the handler
- Added extensions unit tests

Closes #110

### DIFF
--- a/docs/preview/features/message-pumps/customization.md
+++ b/docs/preview/features/message-pumps/customization.md
@@ -75,6 +75,7 @@ public void ConfigureServices(IServiceCollection services)
 ## Fallback message handling
 
 When receiving a message on the message pump and none of the registered `IMessageHandler`'s can correctly process the message, the message pump normally throws and logs an exception.
+
 It could also happen in a scenario that's to be expected that some received messages will not be processed correctly (or you don't want them to).
 
 In such a scenario, you can choose to register a `IFallbackMessageHandler` in the dependency container. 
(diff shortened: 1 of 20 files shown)


tomkerkhove:
Did a quick skim and looks fine (detailed review to follow), but I'm wondering if we could have a variant for SB which gives access to the Service Bus message itself and interact on the `Message` so that we can call abandon() and add context.

Not perse for this PR but just something that came to my mind.

stijnmoreels:
> Did a quick skim and looks fine (detailed review to follow), but I'm wondering if we could have a variant for SB which gives access to the Service Bus message itself and interact on the `Message` so that we can call abandon() and add context.
> 
> Not perse for this PR but just something that came to my mind.

Ah, yes, that would be useful. With the current setup it's not that big a deal to add it. I'll do that.

docs/preview/features/message-pumps/customization.md
```diff
@@ -72,4 +72,40 @@ public void ConfigureServices(IServiceCollection services)
 > Note that the order in which the message handlers are registered is important in the message processing.
 > In the example, when a message handler above this one is registered that could also handle the message (same message type) than that handler may be chosen instead of the one with the specific filter.
 
+## Fallback message handling
+
+When receiving a message on the message pump and none of the registered `IMessageHandler`'s can correctly process the message, the message pump normally throws and logs an exception.
+It could also happen in a scenario that's to be expected that some received messages will not be processed correctly (or you don't want them to).
```
tomkerkhove, line 78: New line for doc rendering
```suggestion

It could also happen in a scenario that's to be expected that some received messages will not be processed correctly (or you don't want them to).
```

tomkerkhove (approved):
LGTM, thanks!

PR #112, opened 2020-09-08
Provide more up-to-date README

tomkerkhove:


stijnmoreels (approved):

stijnmoreels:
Is this something that's in the `index.md` as well? If so, we may need to update the `/preview/index.md`.

tomkerkhove:
We can add it here as well but might be too verbose:

image

Alternative is remove it or simply link to a search for all arcus messaging packages but I'm not sure. I'd leave it as is there given the subpages have the pages I presume?

PR #116, opened 2020-09-18
Fix - rethrow when message handling failed

### LINKED ISSUE #115: Azure Service Bus message pump should rethrow exceptions
**Describe the bug**
Azure Service Bus message pump should rethrow exceptions instead of just logging it:

https://github.com/arcus-azure/arcus.messaging/blob/master/src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs#L135

This is because Azure Service Bus will otherwise complete the failed message when [`AutoComplete`](https://github.com/arcus-azure/arcus.messaging/blob/master/src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpConfiguration.cs#L57) is true.

stijnmoreels:
When the Azure Service Bus message is being handled but fails during processing, the exception should not only be logged, but also be re-thrown so the message is marked abannoned.

Closes #115


tomkerkhove (approved):

tomkerkhove:
Build is stuck, merging anyway.

> Total tests: 9
>      Passed: 9
> Total time: 1.7059 Minutes

PR #124, opened 2020-10-07
Feature - add message body handlers for message processing

### LINKED ISSUE #27: Provide capability to support multiple message handlers based on message body
Provide capability to support multiple message handlers based on message body

This is similar to #26, but purely on the message body stream without any deserialization.

stijnmoreels:
Adds a new registration possibility for `IMessageBodyHandler` interface that allows fine-grained and delegated message deserialization for message processing in the message pump.

Relates to #27


docs/preview/features/message-pumps/customization.md
```diff
@@ -14,6 +14,8 @@ We allow several customizations while implementing your own message pump.
 
 ## Control custom deserialization
 
+### Altering custom message pump
```
tomkerkhove, line 17:
```suggestion
### Extending the message pump
```

docs/preview/features/message-pumps/customization.md
```diff
@@ -14,6 +14,8 @@ We allow several customizations while implementing your own message pump.
```
tomkerkhove, line 14: Can you update this TOC with the subheaders please?

docs/preview/features/message-pumps/customization.md
```diff
@@ -44,6 +46,43 @@ public class OrderMessagePump : MessagePump
...
+
+You start by implemeting an `IMessageBodyHandler`. The following example shows how an expected type can be transformed to something else. 
+The result type (in this case `OrderBatch`) will be then be used to check if there is an `IMessageHandler` registered with that message type.
+
+```csharp
+public class OrderBatchMessageBodyHandler : IMessageBodyHandler
+{
+    public async Task<MessageResult> DeserializeMessageAsync(string messageBody)
+    {
+        var orders = JsonConvert.DeserializeObject<Order[]>(messageBody);
```
tomkerkhove, line 62: Maybe it's more interesting to use an XML example here to more show its use?

docs/preview/features/message-pumps/customization.md
```diff
@@ -44,6 +46,43 @@ public class OrderMessagePump : MessagePump
...
+### Delegating deserialization
+
+You can also choose to delegate the message deserialization to different deserializers. 
+This allows for more reuse, asynchronous, and is available without altering your message pump. 
+
+You start by implemeting an `IMessageBodyHandler`. The following example shows how an expected type can be transformed to something else. 
+The result type (in this case `OrderBatch`) will be then be used to check if there is an `IMessageHandler` registered with that message type.
+
+```csharp
+public class OrderBatchMessageBodyHandler : IMessageBodyHandler
```
tomkerkhove, line 58: Wouldn't it make more sense to emphasize that it's only for serialization?
```suggestion
public class OrderBatchMessageBodySerializer : IMessageBodySerializer
```

docs/preview/features/message-pumps/customization.md
```diff
@@ -44,6 +46,43 @@ public class OrderMessagePump : MessagePump
 }
 ```
 
+### Delegating deserialization
```
tomkerkhove, line 49:
```suggestion
### Bring your own deserialization
```

docs/preview/features/message-pumps/customization.md
```diff
@@ -44,6 +46,43 @@ public class OrderMessagePump : MessagePump
...
+
+You can also choose to delegate the message deserialization to different deserializers.
```
tomkerkhove, line 51:
```suggestion
You can also choose to extend the built-in message deserialization with additional deserializer to meet your needs. 
```

docs/preview/features/message-pumps/customization.md
```diff
@@ -44,6 +46,43 @@ public class OrderMessagePump : MessagePump
 }
 ```
 
+### Delegating deserialization
+
+You can also choose to delegate the message deserialization to different deserializers. 
+This allows for more reuse, asynchronous, and is available without altering your message pump.
```
tomkerkhove, line 52: What about this?
```suggestion
This allows you to easily deserialize into different message formats or reuse existing (de)serialization capabilities that you already have without altering your message pump. 
```

docs/preview/features/message-pumps/customization.md
```diff
@@ -14,6 +14,8 @@ We allow several customizations while implementing your own message pump.
 
 ## Control custom deserialization
```
tomkerkhove, line 15: I would move this after "Filter messages based on message context" & "Fallback message handling" because it will be less used, unless you want to keep this for alphabetically sorting (which is ok)

docs/preview/features/message-pumps/customization.md
```diff
@@ -44,6 +46,43 @@ public class OrderMessagePump : MessagePump
...
+        return MessageResult.Success(new OrderBatch(orders));
+    }
+}
+```
+
+The registration of these message body handlers can be done just as easily as an `IMessageHandler`:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
```
tomkerkhove, line 72: Shouldn't the registration be linked to the message pump that is being registered? Either generally or per message handler because otherwise we will do a lot of deserialization to match things, no?

```csharp
services.AddServiceBusQueueMessagePump("ARCUS_SERVICEBUS_ORDERS_CONNECTIONSTRING")
            .WithServiceBusMessageHandler<OrdersMessageHandler, Order, OrderBatchMessageBodySerializer>();
```

I'm just thinking out loud.

src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -1,16 +1,196 @@
...
     public static class IServiceCollectionExtensions
     {
+        /// <summary>
+        /// Adds a <see cref="IMessageHandler{TMessage}" /> implementation to process the messages from an <see cref="MessagePump"/> implementation.
+        /// resources.
+        /// </summary>
+        /// <typeparam name="TMessageHandler">The type of the implementation.</typeparam>
+        /// <typeparam name="TMessage">The type of the message that the message handler will process.</typeparam>
+        /// <param name="services">The collection of services to use in the application.</param>
+        public static IServiceCollection WithMessageHandler<TMessageHandler, TMessage>(this IServiceCollection services)
```
tomkerkhove, line 23: Do we want to have these general registrations? What is your goal?

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/IMessageBodyHandler.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Threading.Tasks;
+
+namespace Arcus.Messaging.Pumps.Abstractions.MessageHandling
+{
+    /// <summary>
+    /// Represents a handler that provides a deserialization strategy for the incoming message during the message processing of the <see cref="MessagePump"/>.
+    /// </summary>
+    /// <seealso cref="IMessageHandler{TMessage,TMessageContext}"/>
+    /// <seealso cref="MessagePump"/>
+    public interface IMessageBodyHandler
```
tomkerkhove, line 10:
```suggestion
    public interface IMessageBodySerializer
```

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageResult.cs
```diff
@@ -0,0 +1,53 @@
...
+        public static MessageResult Success(object message)
+        {
+            Guard.NotNull(message, nameof(message), "Requires a deserialized message instance when the message deserialization was successful");
+            return new MessageResult(isSuccess: true, result: message);
+        }
+
+        /// <summary>
+        /// Creates a <see cref="MessageResult"/> that represents a faulted deserialization.
+        /// </summary>
+        public static MessageResult Failure()
```
tomkerkhove, line 48: Should we add a message/exception for context?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -216,36 +216,49 @@ protected virtual Task PreProcessMessageAsync<TMessageContext>(MessageHandler me
...
 
-            if (!tryDeserializeToMessageFormat)
+            return false;
+        }
+
+        private async Task<MessageResult> DeserializeMessageAsync(string message, Type messageType)
+        {
+            IEnumerable<IMessageBodyHandler> messageBodyHandlers = ServiceProvider.GetServices<IMessageBodyHandler>();
+
+            foreach (IMessageBodyHandler messageBodyHandler in messageBodyHandlers)
```
tomkerkhove, line 247: I think we should link the custom deserialization to the handler to avoid too many deserialization attempts, certainly for every message

tomkerkhove (commented):
This is a good start but think it needs a bit more tweaks to avoid too many CPU cycles with all the deserialization.

> Closes #27

This is actually not correct, #27 is for determining if a message handler matches a payload by giving users the option to read the raw message and returning an indication if it's a match.

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageResult.cs
stijnmoreels: Yes, thanks.

stijnmoreels:
> This is a good start but think it needs a bit more tweaks to avoid too many CPU cycles with all the deserialization.
> 
> > Closes #27
> 
> This is actually not correct, #27 is for determining if a message handler matches a payload by giving users the option to read the raw message and returning an indication if it's a match.

Hmm, 😅 , I don't really know what's the difference here. I thought that this was what that issue described.

src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: To have independent message determination and message handling. A serializer can be reused across message pumps to have a canonical format for different message handlers.

docs/preview/features/message-pumps/customization.md
stijnmoreels: We could do that, yeah, we should then register the message handler differently but it's possible.

stijnmoreels: Should I move to this approach?

tomkerkhove:
Not really - Here you allow people to control how it is deserialized to the type in general.

However, the goal of the issue was to determine if a message handler is suitable based on the payload. This means they will just give a true/false based on the raw payload, not the concrete message type itself - Let's discuss it in a call if you want.

Think this was your scenario, right @SamVanhoutte?

stijnmoreels:
> Not really - Here you allow people to control how it is deserialized to the type in general.
> However, the goal of the issue was to determine if a message handler is suitable based on the payload. This means they will just give a true/false based on the raw payload, not the concrete message type itself - Let's discuss it in a call if you want.

Aha, so their should an additional predicate that can take in the message body, just like we did with the message context? Right? So, the deserialization itself will remain untouched with this, correct? Only the determination?

stijnmoreels:
I'm wondering if this PR is then even needed, if it's not related to that problem...

tomkerkhove:
I think it does because with this one you can deserialize the messages with something else than JSON while Sam's proposal is just the determination of message handlers by using custom deserialization.

What do you think?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Working on that. Resolving.

src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Resolving as it's changed.

docs/preview/features/message-pumps/customization.md
stijnmoreels: Moved to this approach, resolving.

stijnmoreels:
It's a grand operation here, added more than 280 unit tests to have a more stable message handling method overloading system. It's now more stable to add more functionality in the future and where everything goes.

The message body serializer system is also adapted per message handler, but I was wondering if the result of this serializer should also be typed (as in `IMessageBodySerializer<TMessage>` which returns a `MessageResult<TMessage>`) so we don't have to check for the type that gets deserialized, or is this taking things to far? @tomkerkhove

docs/preview/features/message-pumps/customization.md
```diff
@@ -44,32 +80,43 @@ public class OrderMessagePump : MessagePump
...
 
-We would like that this handler only processed the message when the context contains `MessageType` equals `Order`.
+The registration of these message body handlers can be done just as easily as an `IMessageSerializer`:
 
 ```csharp
 public void ConfigureServices(IServiceCollection services)
 {
-    services.WithMessageHandler<OrderMessageHandler, Order>(context => context.Properties["MessageType"].ToString() == "Order");
+    // Register the message body serializer in the dependency container where the dependent services will be injected.
+    services.WithMessageHandler<OrderBatchMessageHandler>(..., messageBodySerializer: new OrderBatchMessageBodySerializer());
```
tomkerkhove, line 112: Since it's just a new here, why not do this?
```suggestion
    services.WithMessageHandler<OrderBatchMessageHandler, OrderBatchMessageBodySerializer>(...);
```

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -117,7 +117,7 @@ public Type GetMessageHandlerType()
         /// </summary>
         /// <typeparam name="TMessageContext">The type of the message context.</typeparam>
         /// <param name="messageContext">The context in which the incoming message is processed.</param>
-        public bool CanProcessMessage<TMessageContext>(TMessageContext messageContext) where TMessageContext : MessageContext
+        public bool CanProcessMessageBasedOnContext<TMessageContext>(TMessageContext messageContext) where TMessageContext : MessageContext
```
tomkerkhove, line 120: This is abreaking change, can we create a new method and deprecate the old one?

tomkerkhove (approved):
Nice PR, LGMT! 🚀

docs/preview/features/message-pumps/customization.md
stijnmoreels: Because we don't have such an extension 😄 . We could, but it seemed at the time an unnecessary generic. I could add it, but it would mean we create arround 24 extra extensions with all the combinations. (plus tests of course).

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: Well, I don't know if we publicly support this type. It's for internal use only and is only public bc we can use it in our tests and Azure Functions project easily. Not sure if we should do this, but we can of course.

docs/preview/features/message-pumps/customization.md
tomkerkhove: It feels more natural to me, but not worth the effort

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
tomkerkhove: Well if it's public then it's supported🤷‍♂️

We don't know if somebody is extending it so have to support it I'm afraid

stijnmoreels: Voila, done. Resolving.

docs/preview/features/message-pumps/customization.md
stijnmoreels: We could also change the existing ones so that there's a generic overload that creates it directly, and a implementation factory where you get the `IServiceProvider` to create your serializer.

tomkerkhove: Can you give an example please?

stijnmoreels: Now, we have these capabilities:
- `.WithMessageHandler(..., messageBodySerializer: new OrderSerializer());`
- `.WithmessageHandler(..., messageBodySerializerImplementationFactory: serviceProvider => new OrderSerializer());`

We could change it throughout all combinations with the generic instead:
- `.WithMessageHandler<..., OrderSerializer>(...);`
- `.WithmessageHandler(..., messageBodySerializerImplementationFactory: serviceProvider => new OrderSerializer());`

tomkerkhove: I'm confused because isn't that what I suggested 🤔

stijnmoreels: Well, I said we could _add_ also a new overload so that you could both send in the instance as the type and that would lead to 24+ extra extensions, or we could change the existing ones. Not sure what you suggested.

stijnmoreels: Hmm, also, we're not gonna be able to do this approach for every overload bc it will result in some ambigous signatures (same amount of params, only with different generics, still fail).

tomkerkhove: Let's just resolve this for now :)

PR #126, opened 2020-10-14
Provide capability to specify custom transaction id user property to get correlation info

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

Provide capability to specify custom transaction id user property to get correlation info

### DIFF
--- a/src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
+++ b/src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
@@ -35,11 +35,12 @@ namespace Microsoft.Azure.ServiceBus
         ///     Gets the correlation information for a given message.
         /// </summary>
         /// <param name="message">The received message.</param>
-        public static MessageCorrelationInfo GetCorrelationInfo(this Message message)
+        /// <param name="transactionIdPropertyName">Name of the property to determine the transaction id.</param>
+        public static MessageCorrelationInfo GetCorrelationInfo(this Message message, string transactionIdPropertyName = PropertyNames.TransactionId)
         {
             Guard.NotNull(message, nameof(message));
 
-            string transactionId = GetTransactionId(message);
+            string transactionId = GetTransactionId(message, transactionIdPropertyName);
             string operationId = DetermineOperationId(message.CorrelationId);
 
             var messageCorrelationInfo = new MessageCorrelationInfo(operationId, transactionId);
@@ -60,11 +61,12 @@ namespace Microsoft.Azure.ServiceBus
         /// <summary>	
         ///     Gets the transaction id that is linked to this message	
         /// </summary>	
-        /// <param name="message">Message to process</param>	
+        /// <param name="message">Message to process</param>
+        /// <param name="transactionIdPropertyName">Name of the property to determine the transaction id.</param>
         /// <returns>Transaction id for message</returns>
-        public static string GetTransactionId(this Message message)
+        public static string GetTransactionId(this Message message, string transactionIdPropertyName = PropertyNames.TransactionId)
         {
-            return message.UserProperties.TryGetValue(PropertyNames.TransactionId, out object transactionId)
+            return message.UserProperties.TryGetValue(transactionIdPropertyName, out object transactionId)
                 ? transactionId.ToString()
                 : string.Empty;
         }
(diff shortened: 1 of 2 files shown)


src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
```diff
@@ -35,11 +35,12 @@ public static TProperty GetUserProperty<TProperty>(this Message message, string
         ///     Gets the correlation information for a given message.
         /// </summary>
         /// <param name="message">The received message.</param>
-        public static MessageCorrelationInfo GetCorrelationInfo(this Message message)
+        /// <param name="transactionIdPropertyName">Name of the property to determine the transaction id.</param>
+        public static MessageCorrelationInfo GetCorrelationInfo(this Message message, string transactionIdPropertyName = PropertyNames.TransactionId)
```
stijnmoreels, line 39: This should propably be guarded.

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
```diff
@@ -60,11 +61,12 @@ private static string DetermineOperationId(string messageCorrelationId)
         /// <summary>	
         ///     Gets the transaction id that is linked to this message	
         /// </summary>	
-        /// <param name="message">Message to process</param>	
+        /// <param name="message">Message to process</param>
+        /// <param name="transactionIdPropertyName">Name of the property to determine the transaction id.</param>
         /// <returns>Transaction id for message</returns>
-        public static string GetTransactionId(this Message message)
+        public static string GetTransactionId(this Message message, string transactionIdPropertyName = PropertyNames.TransactionId)
```
stijnmoreels, line 67: This should propably be guarded.

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
stijnmoreels, line 39: We'll have to be careful that this doesn't break dependent librarires. When a library only updates this core package and not the other packages, we may get a `MethodNotFound` exception along the way.

src/Arcus.Messaging.Tests.Unit/ServiceBus/MessageExtensionsTests.cs
```diff
@@ -94,6 +94,27 @@ public void GetMessageCorrelationInfo_WithTransactionIdAsUserProperty_ReturnsCor
             Assert.Equal(expectedTransactionId, correlationInfo.TransactionId);
         }
 
+        [Fact]
+        public void GetMessageCorrelationInfo_WithTransactionIdInCustomUserProperty_ReturnsCorrelationInfoWithGeneratedOperationId()
```
stijnmoreels, line 98: An test where we get the custom transaction ID directly can be appropriate also.

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
stijnmoreels, line 39: I'm thinking, is there some place in the docs that this needs to be updated? I think there is.

stijnmoreels (changes requested):

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
tomkerkhove: Doesn't look like it 🤔

src/Arcus.Messaging.Tests.Unit/ServiceBus/MessageExtensionsTests.cs
tomkerkhove: What would that be then? Not sure if I get the suggestion.

src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
tomkerkhove: I've split them!

src/Arcus.Messaging.Tests.Unit/ServiceBus/MessageExtensionsTests.cs
stijnmoreels: Well, there was two methods updated but only one updated method is tested here, right? Or am I wrong?

tomkerkhove: We don't have tests for the other one but will add thme

Issue #132, opened 2020-10-27
Provide correlation information for unhandled exceptions in message handlers


FilipVanRaemdonck:
**Is your feature request related to a problem? Please describe.**
When a custom AzureServiceBusMessageHandler is unable to process a message it throws an error. This error is logged in application insights without correlation information (operationId, transactionId). This makes it hard to correlate the error with other application insights logs.

**Describe the solution you'd like**
The Service Bus Message we couldn't process very likely contains correlation information. It would be good to have these (optionally) in the logs coming from the Arcus Framework.

**Describe alternatives you've considered**
An alternative might be to allow the user to provide a custom error handling function.

tomkerkhove:
Updated the description a bit. Can you please post an example @FilipVanRaemdonck?

@stijnmoreels For some reason, when using correlation enrichment, it does not include the correlation information when tracking exceptions. Now in this case it was related to the interaction with the queue which I think is fair but we'll need to make sure everything is handled correctly.

FilipVanRaemdonck:
![image](https://user-images.githubusercontent.com/14161827/97332117-5ff68400-187a-11eb-98f0-d6bfc009699e.png)

tomkerkhove:
Context for @stijnmoreels, we are looking up secrets as part of the handlers through `ISecretProvider` where it couldn't find one.

stijnmoreels:
OK, and the incoming message contains this correlation information? Could there be a sample of a message and it's properties? Bc the correlation will only be available if the operation and transaction ID is available as properties in the message.

FilipVanRaemdonck:
The correlationId and transactionId are part of the Message properties. The Correlation id is the message's Correlation id, the transactionId is part of the user properties. But you might have to take into account the transactionId name is configurable (https://github.com/arcus-azure/arcus.messaging/pull/126)

tomkerkhove:
@FilipVanRaemdonck can you export a message through Service Bus Explorer (incl. headers) and upload an obfuscated version here please?

FilipVanRaemdonck:
![image](https://user-images.githubusercontent.com/14161827/97415761-e4daaf80-1905-11eb-99d0-7a0f1717544c.png)
Here you are :D

FilipVanRaemdonck:
Would it make sense to put all message custom properties in the logs?

tomkerkhove:
Nope, should be good - Thanks!

stijnmoreels:
[...]

PR #133, opened 2020-10-28
Fix - include correlation information when tracking exceptions

stijnmoreels:
Moved the try/catch/track block inside the correlation information scope.
This wasn't the case which meant that exceptions where tracked without correlation information.

Related to #132

### DIFF
--- a/src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePumpTests.cs
+++ b/src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePumpTests.cs
@@ -214,7 +214,7 @@ namespace Arcus.Messaging.Tests.Integration.MessagePump
             await service.SendMessageToServiceBusAsync(connectionString, orderMessage);
 
             // Assert
-            using (ApplicationInsightsDataClient client = CreateApplicationInsightsClient(applicationInsightsConfig.InstrumentationKey))
+            using (ApplicationInsightsDataClient client = CreateApplicationInsightsClient(applicationInsightsConfig.ApiKey))
             {
                 await RetryAssertUntilTelemetryShouldBeAvailableAsync(async () =>
                 {
@@ -295,7 +295,7 @@ namespace Arcus.Messaging.Tests.Integration.MessagePump
             RetryPolicy retryPolicy =
                 Policy.Handle<Exception>(exception =>
                       {
-                          _logger.LogError(exception, "Failed to contact Azure Application Insights");
+                          _logger.LogError(exception, "Failed to contact Azure Application Insights. Reason: {Message}", exception.Message);
                           return true;
                       })
                       .WaitAndRetryForeverAsync(index => TimeSpan.FromSeconds(1));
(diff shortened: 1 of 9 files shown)


stijnmoreels:
Question though, normally we would write a test to prove the fix of the issue, but in this case, we would have to run a message handler via the message pump, sabotage the processing so an exception is thrown, send it via application insights and verify if the correlation information is one of the properties there. Is that worth the effort?

stijnmoreels:
Also @tomkerkhove , we have allowed to set a custom transaction ID message property, but right now, you can't change it when using the Azure Service Bus message pump. We may want to allign the configuration settings of this pump with the custom transaction name (default the one we're using now), so consumers can change this, right?

stijnmoreels:
> Question though, normally we would write a test to prove the fix of the issue, but in this case, we would have to run a message handler via the message pump, sabotage the processing so an exception is thrown, send it via application insights and verify if the correlation information is one of the properties there. Is that worth the effort?

I can do that of course if you like.

tomkerkhove:
That's a good idea, thanks @stijnmoreels!

We can indeed send it to AI as I think we should have those checks but maybe as a dedicated item given it's pretty important, no?

src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePumpTests.cs
```diff
@@ -185,6 +191,48 @@ await using (var service = await TestMessagePumpService.StartNewAsync(config, _l
...
+            using (ApplicationInsightsDataClient client = CreateApplicationInsightsClient(applicationInsightsConfig.InstrumentationKey))
+            {
+                await RetryAssertUntilTelemetryShouldBeAvailableAsync(async () =>
+                {
+                    const string onlyLastHourFilter = "timestamp gt now() sub duration'PT1H'";
+                    EventsResults<EventsExceptionResult> results = 
+                        await client.Events.GetExceptionEventsAsync(applicationInsightsConfig.ApplicationId, filter: onlyLastHourFilter);
+
+                    Assert.Contains(results.Value, result =>
+                    {
```
tomkerkhove, line 226: Can we also assert that the operation & transaction id are assigned to the correct AI fields, such as `operation.id` for operation id?

This requires our sink of course.

tomkerkhove (approved):

src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePumpTests.cs
stijnmoreels: Yeah, we can do that. We're already using our sink here, so that's OK.

stijnmoreels: Voila ✔️

stijnmoreels: The transaction ID not yet set, as we've discussed.

tomkerkhove: 💘

PR #135, opened 2020-11-06
Fix - combine context predicate w/ Service Bus operations

### LINKED ISSUE #134: Provide capability to combine message context filtering and message operations for Azure Service Bus
Provide the capability to combine message context filtering and message operations for Azure Service Bus which is not supported today.

When using message context filtering as following:

```csharp
services.AddServiceBusTopicMessagePump(topicName, subscriptionName, secretProvider => secretProvider.GetRawSecretAsync("ServiceBus_ConnectionString"), options => options.TopicSubscription = TopicSubscription.None)
                .WithServiceBusMessageHandler<DeviceRegistrationV2MessageHandler, DeviceRegistrationV2Message>(messageContext => messageContext.Properties["MessageType"] == MessageType.DeviceRegistrationV2.ToString());
```

Then you cannot use operations such as `AbandonMessageAsync` with `AzureServiceBusMessageHandler`.

stijnmoreels:
The problem existed when we tried to use both the context filtering as the Azure Service Bus operations when handling a message.
By hiding how we determine the type and instance of the message handler, we're more in control and can cirvumvent this problem.

Closes #134

### DIFF
--- a/src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
+++ b/src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/MessageHandler.cs
@@ -18,6 +18,8 @@ namespace Arcus.Messaging.Pumps.Abstractions.MessageHandling
     /// </summary>
     public class MessageHandler
     {
+        private readonly object _service;
+        private readonly Type _serviceType;
         private readonly ILogger _logger;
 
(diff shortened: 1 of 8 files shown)


stijnmoreels:
Some problem with connecting with Application Insights, not related to the changes made here. Will look at it.

stijnmoreels:
Voila, fixed it.

PR #138, opened 2020-11-20
Feature - add capability to route message based on message body

### LINKED ISSUE #27: Provide capability to support multiple message handlers based on message body

stijnmoreels:
Provide the capability to route incoming messages not only on the message context in which the message is processed, but on the non-deserialized message content itself.
This works the same way as the context filtering works in the message handling, only now you can pass along one or the other or both.

Closes #27


docs/preview/features/message-pumps/customization.md
```diff
@@ -76,6 +76,44 @@ public void ConfigureServices(IServiceCollection services)
...
+}
+
+// Message handler registration
+public class Startup
+{
+    ...
+    
+    public void ConfigureServices(IServiceCollection services)
+    {
+        services.WithMessageHandler<OrderMessageHandler, Order>(body => body.Contains("Done"));
```
tomkerkhove, line 112: Maybe we should expand this sample so that it uses JToken.Parse and uses token["status"] == "Done"? (or even change Status type to an enum OrderStatus)

tomkerkhove: Adding @SamVanhoutte as he initially requested this feature

tomkerkhove (approved):
LGTM but will wait for @SamVanhoutte

docs/preview/features/message-pumps/customization.md
SamVanhoutte: Looks good to me.  Agree having a sample to subscribe on json-payload would be great.

SamVanhoutte (approved):

docs/preview/features/message-pumps/customization.md
stijnmoreels: I'm wondering if we should use the typed-predicate here, as in `(Order order) => order.Status == OrderStatus.Done`). Internally, it's not that big a change and it eliminates the 'risk' that consumers have the parse the raw body themselves?

tomkerkhove: Fine for me!

stijnmoreels: Should we then still support the raw string predicate? Or only the typed predicate?

tomkerkhove: If it's introduced in this PR then we can remove it, otherwise we have to still support it

PR #141, opened 2020-12-16
feat: extract message routing from the message pump

### LINKED ISSUE #127: Provide message handler support for Azure Functions
Provide message handler support for Azure Functions which is similar to our message pump scenario, but then by using Azure Function Service Bus bindings.

This would allow us to use Azure Functions as the message pump, but still use message handlers with it.

Let's work on an API design together on this based on a POC which can become a template as well then.

stijnmoreels:
Extracts the message routing that includes the selection of the correct registered message handler for the incoming message to be processed.

Because the `MessagePump` and by result the `AzureServiceBusMessagePump` already contained methods like `ProcessMessageAsync` which handles this message routing, we can only mark them as obsolete but can't change them in any way.

This PR includes the introduction of the `IMessageRouter` and `IAzureServiceBusMessageRouter` which is the exact functionality of this message routing. The `AzureServiceBusMessagePump` already uses this router so the existing tests will suffice to make sure that this routing is still valid. An extra interface for the Service Bus was required, because the message handling required specific Service Bus operations calls and the original Service Bus message upon fallback.

Implementation of POC: https://github.com/stijnmoreels/arcus.messaging.azurefunctions
Relates to #127


stijnmoreels:
Currently WIP, TODO list:
- [x] Message routing implementation
- [x] Unit tests to proof message routing implementation
- [ ] Feature docs update to specify how consumers can use this routing in their own projects (ex. Azure Functions).

docs/preview/features/message-pumps/service-bus-azure-functions.md
```diff
@@ -0,0 +1,150 @@
...
+[assembly: FunctionsStartup(typeof(Startup))]
+namespace MessageProcessing
+{
+    public class Startup : FunctionsStartup
+    {
+        public override void Configure(IFunctionsHostBuilder builder)
+        {
+            builder.Services.AddServiceBusMessageRouting()
+                            .WithServiceBusMessageHandler<OrderServiceBusMessageHandler, Order>(messageContext => messageContext.UserProperties["MessageType"] == MessageType.Order)
+                            .WithServiceBusMessageHandler<ShipmentServiceBusMessageHandler, Shipment>(messageContext => messageContext.UserProperties["MessageType"] == MessageType.Shipment);
```
stijnmoreels, line 113: We could consider to extract these extensions to the seperate 'core' project of Service Bus and create some new extensions for the specific Azure Functions `IFunctionHostBuilder` type. But not sure if this is 
1. worth the effort, since it will be first of all require some work to duplicate all possible combinations to this new type
2. all namespaces will be different then. Message handler interfaces are now in the `...Pumps.ServiceBus` namespace while they will be moved to `...ServiceBus.Core`. We could of course let them exist in this 'old' namespace so we don't break anything; and move it in a later stage.

stijnmoreels:
I've not updated the `MessagePump` (yet) with using the `IMessageRouter` instead, because that would mean that we require such a registration in the `IServiceProvider`; which would mean that we fail when there's no such registration. And this is actually a breaking change, so I've let the `protected` methods on the pump as-is but marked them all obsolete until we can safely in a new major version break this.

src/Arcus.Messaging.Pumps.ServiceBus/MessageHandling/IAzureServiceBusMessageRouter.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// </remarks>
+        /// <exception cref="ArgumentNullException">
+        ///     Thrown when the <paramref name="message"/>, <paramref name="messageContext"/>, or <paramref name="correlationInfo"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="InvalidOperationException">Thrown when no message handlers or none matching message handlers are found to process the message.</exception>
+        Task ProcessMessageAsync(
+            Message message,
+            AzureServiceBusMessageContext messageContext,
+            MessageCorrelationInfo correlationInfo,
+            CancellationToken cancellationToken);
```
stijnmoreels, line 36: You could maybe argue that we only should allow an `ProcessMessageAsync` overload with a `MessageReceiver`, but that felt a bit pushing upon the consumer. If you don't uses any of the Service Bus specific operations, you should maybe not be forced to pass along this argument.

stijnmoreels: Just like you're also free to use the 'general' overload that disregard the service bus-specific fallback handler.

docs/preview/features/message-pumps/service-bus-azure-functions.md
stijnmoreels: Aha, there's also option 3. where we create a separate extension for message routing for `IFunctionHostBuilder` but return an `IServiceCollection` instead so we can reuse the other message handler extensionns.

stijnmoreels: This last option is maybe the best, because otherwise we would have to include the message pump library in a Azure Functions project, while we only would need the message handlers in that case.

So, proposal to move the message handlers to a the ServiceBus Core project so we can reuse it in both the message pump project and Azure Functions project. The problem I see with this approach, is that the interface for the message handler also would have to move to the Core project and that would be a breaking change.

tomkerkhove: Since we have a specific package already I think we should only provide extensions that use IFunctionHostBuilder as input and call the others under the hood, to be honest; or even introduce a service specific Azure function package (my pref).

Better do it right now than think it would have been better later on :D

docs/preview/features/message-pumps/service-bus-azure-functions.md
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+}
+```
+
+This extension will register an `IAzureServiceBusMessageRouter` interface allows you access to message handling with specific Service Bus operations during the message processing (like dead lettering and abandonning).
+It also registers an more general `IMessageRouter` you can use if the general message routing (with the message raw message body as `string` as incoming message) will suffice.
```
tomkerkhove, line 120:
```suggestion

It also registers an more general `IMessageRouter` you can use if the general message routing (with the message raw message body as `string` as incoming message) will suffice.
```

docs/preview/features/message-pumps/service-bus-azure-functions.md
```diff
@@ -0,0 +1,150 @@
...
+
+In any case, this router can be injected in your function:
```
tomkerkhove, line 122: We can now inject the message router in our Azure Function and process all messages with it.

This will determine what the matching message handler is and process it accordingly:

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/IMessageRouter.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// </summary>
+        /// <param name="message">The message that was received.</param>
+        /// <param name="messageContext">The context providing more information concerning the processing.</param>
+        /// <param name="correlationInfo">The information concerning correlation of telemetry and processes by using a variety of unique identifiers.</param>
+        /// <param name="cancellationToken">The token to cancel the message processing.</param>
+        /// <exception cref="ArgumentNullException">
+        ///     Thrown when the <paramref name="message"/>, <paramref name="messageContext"/>, or <paramref name="correlationInfo"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="InvalidOperationException">Thrown when no message handlers or none matching message handlers are found to process the message.</exception>
+        Task ProcessMessageAsync<TMessageContext>(
```
tomkerkhove, line 25: If it has a process, is it then still a router? :)

stijnmoreels: Maybe rename to 'route' or something? We should come up with something here, as it's a new interface.

docs/preview/features/message-pumps/service-bus-azure-functions.md
stijnmoreels: Updated that line with your suggestion, thanks! Resolving.

docs/preview/features/message-pumps/customization.md
```diff
@@ -56,10 +56,22 @@ public class Startup
...
+By default, when using the `.AddMessageRouting`, an new `IMessageRouter` instance is registered that contains the basic message routing/handling for incoming messages.
+If you want for some reason alter this routing process or provide additional functionality, you can do this by implementing your own message router. You can also inherit from our built-in message router and override the necessary methods of your choosing.
+
+Here's how you normally would register the message router:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    // Registers `IMessageRouter`.
+    services.AddMessageRouting();
```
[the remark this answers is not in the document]
stijnmoreels: > Just curious - If I don't want to extend it, should this be called?

Yes, then this should be called.

> If so, can't we do this as part of the message pump registration?

Yes, and our Azure Service Bus router is already registered via registering the Azure Service Bus message pump.
The abstract message pump can't be registered by itself, so it has to be implemented either way. When you want to use the message routing functionality in that message pump registration, you should use the message routing extension. But, like we did with our Azure Service Bus message pump, this is done behind the scenes.

docs/preview/features/message-pumps/service-bus.md
```diff
@@ -352,6 +353,52 @@ public class Startup
...
+    }
+}
+```
+
+This custom message router can be registered with the following extension:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddServiceBusMessageRouting(serviceProvider =>
```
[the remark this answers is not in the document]
stijnmoreels: That's not needed, bc the routing is done while registering the Azure Service Bus message pump.

docs/preview/features/message-pumps/service-bus-azure-functions.md
stijnmoreels: Ok, hmm, then we should add 64+ extensions which are almost exact duplicates, so we can use this interface instead.

stijnmoreels: The question here was also whether or not to move the message handling interfaces as they are both used in message pumps as Azure Functions projects.

docs/preview/features/message-pumps/service-bus.md
```diff
@@ -352,6 +353,52 @@ public class Startup
 }
 ```
 
+## Alternative Service Bus message routing
+
+By deafult, when registering the Azure Service Bus message pump with `AddServiceBusXxxMessagePump`, a built-in message router is registered to handle the routing throughout the previously registered message handlers.
+This router is registered with the `IAzureServiceBusMessageRouter` interface (which implements the more general `IMessageRouter` for non-specific Service Bus messages).
```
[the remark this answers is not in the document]
stijnmoreels: Yes, I see your point. Maybe this remark can be moved to a code remark instead of something found in the documentation.

docs/preview/features/message-pumps/service-bus-azure-functions.md
```diff
@@ -0,0 +1,150 @@
...
+While our default message pump system provides a way to receive, route and handle incoming Service Bus messages; in Azure Functions this is a whole other story.
+An Azure Function that gets triggered upon a Service Bus message already has a system to receive such a message and only has to route and handle the message. The function functions as a 'message pump' in this scenario.
+
+That's why we extracted our message routing functionality so you can call it directly from your Azure Function.
+
+- [Receive Azure Service Bus message in an Azure Function](#receive-azure-service-bus-message-in-an-azure-function)
+- [Determine Service Bus message handlers](#determine-service-bus-message-handlers)
+- [Activate message routing](#activate-message-routing)
+
+## Receive Azure Service Bus message in an Azure Function
```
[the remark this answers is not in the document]
stijnmoreels: I'll resolve this.

docs/preview/features/message-pumps/service-bus-azure-functions.md
stijnmoreels: > Ok, hmm, then we should add 64+ extensions which are almost exact duplicates, so we can use this interface instead.

Definitely in another issue I would say, if we go through with this.

src/Arcus.Messaging.Pumps.Abstractions/MessageHandling/IMessageRouter.cs
stijnmoreels: I'll update it to `RouteMessageAsync`.

tomkerkhove:
> I've not updated the `MessagePump` (yet) with using the `IMessageRouter` instead, because that would mean that we require such a registration in the `IServiceProvider`; which would mean that we fail when there's no such registration. And this is actually a breaking change, so I've let the `protected` methods on the pump as-is but marked them all obsolete until we can safely in a new major version break this.

Do we have an issue to improve this?

tomkerkhove:
Would it be easier if you could make breaking changes and we move this to v1.0?

stijnmoreels:
> > I've not updated the `MessagePump` (yet) with using the `IMessageRouter` instead, because that would mean that we require such a registration in the `IServiceProvider`; which would mean that we fail when there's no such registration. And this is actually a breaking change, so I've let the `protected` methods on the pump as-is but marked them all obsolete until we can safely in a new major version break this.
> 
> Do we have an issue to improve this?

No, not yet.

stijnmoreels:
> Would it be easier if you could make breaking changes and we move this to v1.0?

Hmm, yeah, maybe. If we would do that, I would also move the `IMessageHandler` to the `Messaging.Abstractions` project and move the `IAzureServiceBusMessageHandler` to the `Messaging.ServiceBus.Core` project (or a new project called `.Abstractions`) so we can add extensions for the Azure Functions type without referencing any message pump projects..

src/Arcus.Messaging.Pumps.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
```diff
@@ -0,0 +1,260 @@
...
+                Encoding encoding = messageContext.GetMessageEncodingProperty(Logger);
+                string messageBody = encoding.GetString(message.Body);
+
+                foreach (MessageHandler messageHandler in messageHandlers)
+                {
+                    MessageResult result = await DeserializeMessageForHandlerAsync(messageBody, messageContext, messageHandler);
+                    if (result.IsSuccess)
+                    {
+                        SetServiceBusPropertiesForSpecificOperations(messageHandler, messageReceiver, messageContext);
+                        await messageHandler.ProcessMessageAsync(result.DeserializedMessage, messageContext, correlationInfo, cancellationToken);
```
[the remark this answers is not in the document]
tomkerkhove: What do you think @fgheysels?

docs/preview/features/message-pumps/service-bus-azure-functions.md
tomkerkhove: > Ok, hmm, then we should add 64+ extensions which are almost exact duplicates, so we can use this interface instead.

It's ok to leave out for now then, it just makes it less discoverable :)

> The question here was also whether or not to move the message handling interfaces as they are both used in message pumps as Azure Functions projects.

Sharing is ok for me :)

src/Arcus.Messaging.Pumps.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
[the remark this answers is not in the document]
tomkerkhove: I'm ok with it, but just don't want to have to rename it later on again :)

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -95,6 +93,7 @@ protected virtual Task HandleReceiveExceptionAsync(Exception receiveException)
         /// </summary>
         /// <param name="messageHandler">The message handler to be used to process the message.</param>
         /// <param name="messageContext">The message context of the message that will be handled.</param>
+        [Obsolete("Consider using the " + nameof(IServiceCollectionExtensions.AddMessageRouting) + " instead to register an " + nameof(IMessageRouter) + " implementation that you can use directly in your message pump")]
```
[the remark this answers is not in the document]
stijnmoreels: The `MessagePump` is an abstraction (`abstract`) so can't be added to the application by itself. It has to be added via a dependent class (like `AzureServiceBusMessagePump`). While registering the Azure Service Bus message pump, we can behind the scenes register our router, but since we don't have this opportunity with the `abstract` `MessagePump`, we can't do that. These `[Obsolete]` methods can still be used though, if you get (lazy) our `IMessageRouter` from the services.

stijnmoreels:
> Hm then I'm confused - Do I always have to call AddMessageRouting or not? Because you've mentioned that the message pump registration handles it for us; or is it only for Azure Service Bus?

Yes, we have to call it always, but it's being called behind the scenes with the Azure Service Bus message pump registration.

src/Arcus.Messaging.Pumps.ServiceBus/MessageHandling/IAzureServiceBusFallbackMessageHandler.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.Threading;
 using System.Threading.Tasks;
 using Arcus.Messaging.Abstractions;
+using Arcus.Messaging.Pumps.Abstractions.MessageHandling;
```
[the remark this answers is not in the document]
stijnmoreels: 😄 , no indeed.

docs/preview/features/message-pumps/service-bus-azure-functions.md
```diff
@@ -0,0 +1,158 @@
...
+Today, Azure Functions acts as a message receiver meaning that the function is triggered when a message is available but does not handle message routing and handling. However, in this case, it acts as the message pump.
+
+That's why we extracted our message routing functionality so you can call it directly from your Azure Function.
+
+We will walk you through the process of using message handlers with Azure Functions:
+
+1. [Create a new Azure Function with Service Bus binding](#receive-azure-service-bus-message-in-an-azure-function)
+2. [Declaring our Azure Service Bus message handlers](#specifying-service-bus-message-handlers)
+3. [Enabling message routing](#using-message-routing)
+4. Processing received messages through the message router
```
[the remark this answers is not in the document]
stijnmoreels: Done, resolving.

docs/preview/features/message-pumps/service-bus-azure-functions.md
```diff
@@ -0,0 +1,158 @@
...
+While our default message pump system provides a way to receive, route, and handle incoming Service Bus messages which are, unfortunately, not supported in Azure Functions.
+Today, Azure Functions acts as a message receiver meaning that the function is triggered when a message is available but does not handle message routing and handling. However, in this case, it acts as the message pump.
+
+That's why we extracted our message routing functionality so you can call it directly from your Azure Function.
+
+We will walk you through the process of using message handlers with Azure Functions:
+
+1. [Create a new Azure Function with Service Bus binding](#receive-azure-service-bus-message-in-an-azure-function)
+2. [Declaring our Azure Service Bus message handlers](#specifying-service-bus-message-handlers)
+3. [Enabling message routing](#using-message-routing)
```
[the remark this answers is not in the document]
stijnmoreels: Done, resolving.

src/Arcus.Messaging.Pumps.ServiceBus/MessageHandling/IAzureServiceBusFallbackMessageHandler.cs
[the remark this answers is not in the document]
stijnmoreels: Removed, resolving. Thanks!

tomkerkhove:
> > Hm then I'm confused - Do I always have to call AddMessageRouting or not? Because you've mentioned that the message pump registration handles it for us; or is it only for Azure Service Bus?
> 
> Yes, we have to call it always, but it's being called behind the scenes with the Azure Service Bus message pump registration.

But since we don't have any others, is it useful to mention that even? Isn't it confusing?

tomkerkhove:
> > > I've not updated the `MessagePump` (yet) with using the `IMessageRouter` instead, because that would mean that we require such a registration in the `IServiceProvider`; which would mean that we fail when there's no such registration. And this is actually a breaking change, so I've let the `protected` methods on the pump as-is but marked them all obsolete until we can safely in a new major version break this.
> > 
> > 
> > Do we have an issue to improve this?
> 
> No, not yet.

Let's add it

tomkerkhove:
> > Would it be easier if you could make breaking changes and we move this to v1.0?
> 
> Hmm, yeah, maybe. If we would do that, I would also move the `IMessageHandler` to the `Messaging.Abstractions` project and move the `IAzureServiceBusMessageHandler` to the `Messaging.ServiceBus.Core` project (or a new project called `.Abstractions`) so we can add extensions for the Azure Functions type without referencing any message pump projects..

Let's target 1.0 for this then?

stijnmoreels:
> > > Hm then I'm confused - Do I always have to call AddMessageRouting or not? Because you've mentioned that the message pump registration handles it for us; or is it only for Azure Service Bus?
> > 
> > 
> > Yes, we have to call it always, but it's being called behind the scenes with the Azure Service Bus message pump registration.
> 
> But since we don't have any others, is it useful to mention that even? Isn't it confusing?

Yeah, I see your point. Maybe we should remove this and let consumer see it when they implement their own message pump.

stijnmoreels:
> > > Would it be easier if you could make breaking changes and we move this to v1.0?
> > 
> > 
> > Hmm, yeah, maybe. If we would do that, I would also move the `IMessageHandler` to the `Messaging.Abstractions` project and move the `IAzureServiceBusMessageHandler` to the `Messaging.ServiceBus.Core` project (or a new project called `.Abstractions`) so we can add extensions for the Azure Functions type without referencing any message pump projects..
> 
> Let's target 1.0 for this then?

Yes, OK, but let's put the these breaking namespace moves in another PR.
I'll make an issue for that: https://github.com/arcus-azure/arcus.messaging/issues/153

stijnmoreels:
OK, after reading all the remaining comments, there's only 2 that remains unsolved, and that's the mentioning of the `.AddMessageRouting` in the **Customization** feature docs (which we will discuss offline).

And the question how this extraction should be called 'Router' or 'Processor' or ... (honestly, if it were 'Processor' it would be confusing in combination with the handlers. "Process" <> "Handle" message?)

tomkerkhove:
Let's stick with Router then.

> OK, after reading all the remaining comments, there's only 2 that remains unsolved, and that's the mentioning of the .AddMessageRouting in the Customization feature docs (which we will discuss offline).

What is this topic again?

stijnmoreels:
> Let's stick with Router then.
> 
> > OK, after reading all the remaining comments, there's only 2 that remains unsolved, and that's the mentioning of the .AddMessageRouting in the Customization feature docs (which we will discuss offline).
> 
> What is this topic again?

Described here: https://github.com/arcus-azure/arcus.messaging/pull/141/files#diff-a29b66ed26917dda14cce134994c1fd15605f0ecfbc72dd180eb355be4ee0b7cR71
Under the topic `Extending the Message Router`
Altough it seems reasonable to mention this if we're talking about extending and customization. It's already an advanced topic.

docs/preview/features/message-pumps/service-bus-azure-functions.md
stijnmoreels: Added an issue for this: #154 . Resolving.

tomkerkhove:
> > Let's stick with Router then.
> > > OK, after reading all the remaining comments, there's only 2 that remains unsolved, and that's the mentioning of the .AddMessageRouting in the Customization feature docs (which we will discuss offline).
> > 
> > 
> > What is this topic again?
> 
> Described here: [#141 (files)](https://github.com/arcus-azure/arcus.messaging/pull/141/files#diff-a29b66ed26917dda14cce134994c1fd15605f0ecfbc72dd180eb355be4ee0b7cR71)
> Under the topic `Extending the Message Router`
> Altough it seems reasonable to mention this if we're talking about extending and customization. It's already an advanced topic.

Sure thing, but it's pending a reply from you, no?

stijnmoreels:
> > > Let's stick with Router then.
> > > > OK, after reading all the remaining comments, there's only 2 that remains unsolved, and that's the mentioning of the .AddMessageRouting in the Customization feature docs (which we will discuss offline).
> > > 
> > > 
> > > What is this topic again?
> > 
> > 
> > Described here: [#141 (files)](https://github.com/arcus-azure/arcus.messaging/pull/141/files#diff-a29b66ed26917dda14cce134994c1fd15605f0ecfbc72dd180eb355be4ee0b7cR71)
> > Under the topic `Extending the Message Router`
> > Altough it seems reasonable to mention this if we're talking about extending and customization. It's already an advanced topic.
> 
> Sure thing, but it's pending a reply from you, no?

No, it's anwsered here: https://github.com/arcus-azure/arcus.messaging/pull/141#issuecomment-779731730, but can be disussed offline.

stijnmoreels:
> > Hm then I'm confused - Do I always have to call AddMessageRouting or not? Because you've mentioned that the message pump registration handles it for us; or is it only for Azure Service Bus?
> 
> Yes, we have to call it always, but it's being called behind the scenes with the Azure Service Bus message pump registration.

This was the latest awnsered question. But maybe you'll need to revisit the PR to know what's going on again 😄 .

tomkerkhove (approved):

tomkerkhove:
Super helpful, thanks!

tomkerkhove:
I see the PR closes the issue, did we add the handler for Functions as well? Think I saw it but want to be sure.

stijnmoreels:
> I see the PR closes the issue, did we add the handler for Functions as well? Think I saw it but want to be sure.

We can include that in #154 I guess. But I'll change it bc it's technically indeed not completely done. ☺️

PR #142, opened 2020-12-30
Feature - log security event upon Azure Service Bus connection string retrieval

stijnmoreels:
Make sure that we track the retrieval of the Azure Service Bus connection string when the message pump requests this.

### DIFF
--- a/src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
+++ b/src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
@@ -6,6 +6,8 @@ using Arcus.Security.Core.Caching;
 using GuardNet;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Arcus.Messaging.Pumps.ServiceBus.Configuration
 {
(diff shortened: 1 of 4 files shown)


src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
```diff
@@ -81,6 +83,17 @@ public class AzureServiceBusMessagePumpSettings
         /// <returns>Connection string to authenticate with</returns>
         public async Task<string> GetConnectionStringAsync()
         {
+            ILogger logger = 
+                _serviceProvider.GetService<ILogger<AzureServiceBusMessagePumpSettings>>() 
+                ?? NullLogger<AzureServiceBusMessagePumpSettings>.Instance;
+
+            logger.LogSecurityEvent("Get Azure Service Bus connection string", new Dictionary<string, object>
```
tomkerkhove, line 90: Please refresh my memory; will it track security events out-of-the-box or is this an opt-in?

stijnmoreels: Hmm, not, the option to opt-out for security events is something we added to the secret store functionality in the security repo. We didn't had security events logging in the messaging repo, and therefore no opt-out functionality.

tomkerkhove: I would add it then because by default I would prefer not to have these since they can pollute our telemetry

stijnmoreels: OK, done. Added options to opt-in for security events.

tomkerkhove (approved):

PR #157, opened 2021-02-25
Feature - make message correlation transaction ID configurable

### LINKED ISSUE #156: Provide capability to configure correlation property names in message pump
Provide the capability to configure correlation property names in message pump.

You can do it when using the `GetCorrelationInfo` extension, but this is used in the pump where we cannot access it: https://github.com/arcus-azure/arcus.messaging/blob/master/src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs#L471

stijnmoreels:
Introduces a new property in the settings which can control the name of correlation transaction ID of the Azure Service Bus message during receival of Azure Service Bus messages in the message pump.

Closes #156

### DIFF
--- a/docs/preview/features/message-pumps/service-bus.md
+++ b/docs/preview/features/message-pumps/service-bus.md
@@ -167,6 +167,10 @@ public class Startup
                 // this job instance in a multi-instance deployment (default: guid).
                 options.JobId = Guid.NewGuid().ToString();
 
+                // The name of the Azure Service Bus message property that has the transaction ID.
+                // (default: Transaction-Id).
+                options.TransactionIdPropertyName = "X-Transaction-ID";
+
                 // Indicate whether or not a new Azure Service Bus Topic subscription should be created/deleted
                 // when the message pump starts/stops (default: CreateOnStart & DeleteOnStop).
                 options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
--- a/docs/preview/features/message-pumps/service-bus.md
+++ b/docs/preview/features/message-pumps/service-bus.md
@@ -169,7 +169,7 @@ public class Startup
 
                 // The name of the Azure Service Bus message property that has the transaction ID.
                 // (default: Transaction-Id).
-                options.TransactionIdPropertyName = "X-Transaction-ID";
+                options.Correlation.TransactionIdPropertyName = "X-Transaction-ID";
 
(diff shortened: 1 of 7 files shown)


docs/preview/features/message-pumps/service-bus.md
```diff
@@ -167,6 +167,10 @@ public class Startup
                 // this job instance in a multi-instance deployment (default: guid).
                 options.JobId = Guid.NewGuid().ToString();
 
+                // The name of the Azure Service Bus message property that has the transaction ID.
+                // (default: Transaction-Id).
+                options.TransactionIdPropertyName = "X-Transaction-ID";
```
tomkerkhove, line 172: What about this?
```suggestion
                options.Correlation.TransactionIdPropertyName = "X-Transaction-ID";
```

Because I'm also considering adding a flag where a new transaction id is generated in case it's not specified on the message. Thoughts on that?

stijnmoreels: Yes, that seems like a good approach, similar as we did in the web API repo. I'll change that!

stijnmoreels: Updated this. Come and see! 😉

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -468,7 +468,7 @@ private async Task HandleMessageAsync(Message message, CancellationToken cancell
                 Logger.LogTrace("No operation ID was found on the message");
             }
 
-            MessageCorrelationInfo correlationInfo = message.GetCorrelationInfo();
+            MessageCorrelationInfo correlationInfo = message.GetCorrelationInfo(Settings.Options.Correlation.TransactionIdPropertyName);
```
tomkerkhove, line 471: What if it's null? Are we sure this is safe?

tomkerkhove (approved):

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: It's should be checked when creating the options, but we can always add some paranoia checks if you like :)

tomkerkhove (approved):

PR #160, opened 2021-02-26
Feature - initialize the transaction ID (like the operation ID) with default guid

### LINKED ISSUE #158: Provide capability to initialize transaction id in message pump when not provided on message
Provide capability to initialize transaction id in message pump when not provided on message so that a transaction id is always used.

stijnmoreels:
Update the message correlation extraction to initialize like the operation ID the transaction ID with a default guid when none is found.

Closes #158

### DIFF
--- a/src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
+++ b/src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
@@ -52,11 +52,23 @@ namespace Microsoft.Azure.ServiceBus
             Guard.NotNull(message, nameof(message));
             Guard.NotNullOrWhitespace(transactionIdPropertyName, nameof(transactionIdPropertyName));
 
-            string transactionId = GetTransactionId(message, transactionIdPropertyName);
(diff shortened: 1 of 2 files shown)


src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
```diff
@@ -93,9 +105,12 @@ public static string GetTransactionId(this Message message, string transactionId
...
 
-            return message.UserProperties.TryGetValue(transactionIdPropertyName, out object transactionId)
-                ? transactionId.ToString()
-                : string.Empty;
+            if (message.UserProperties.TryGetValue(transactionIdPropertyName, out object transactionId))
+            {
+                return transactionId.ToString();
+            }
+
+            return string.Empty;
```
stijnmoreels, line 113: Not really a fan returning an empty string here, can this be `null` instead?

tomkerkhove: Sure

tomkerkhove (approved):

PR #166, opened 2021-03-23
fix: stabilize integration tests

stijnmoreels:
Stabilize Azure Service Bus message pump integration tests by stepping down from the approach that started/stopped an entire project to instead host the message pump directly in the running test project process. This allows for much easier disposal of resources and dependency control, plus allows us to be more descriptive in the test itself.
By using this approach logging will become also easier when a test has problems because we can redirect directly the message pump logging to the test output.


stijnmoreels:
We're getting there! Everything works now, only refactoring, cleanup, and some restyling and we're good to go for review, and then release!

tomkerkhove (approved):

PR #169, opened 2021-03-31
chore(rename): use built-in 'ServiceBusEntityType' for our message pump config

### LINKED ISSUE #106: Rename `ServiceBusEntity` to `ServiceBusEntityType`
Rename `ServiceBusEntity` to `ServiceBusEntityType` which is more clear what it does.

stijnmoreels:
We'll use the built-in Azure Service Bus entity type for better understanding.

Closes #106


stijnmoreels:
Hmm, got a reference unit/integration test exception bc the key rotation project uses the cloud event background jobs; which internally still uses our v0.1.0 messaging package...

tomkerkhove (approved):

PR #172, opened 2021-04-09
feat: fluent API discovery with dedicated type for message handling

### LINKED ISSUE #152: Improve fluent API with builder pattern when registering Azure Service Bus message pump
Improve fluent API with builder pattern so that we control what the following methods are that can be called.

```csharp
services.AddServiceBusQueueMessagePump(secretName, options => options.JobId = jobId)
            .WithServiceBusMessageHandler<OrdersAzureServiceBusMessageHandler, Order>();
services.WithAutoRestartServiceBusQueueMessagePumpOnRotatedCredentials(...);
```

This is to make it more clear and less verbose since every extension we have today is returning `IServiceCollection`.

This also helps new people to explore what the other options are for configuring the runtime.

Return a `IServiceBusMessagePumpBuilder` which enforces you to call `WithServiceBusMessageHandler`, `WithAutoRestartOnRotatedCredentials` and/or ... after `AddServiceBusQueueMessagePump`

```diff
services.AddServiceBusQueueMessagePump(secretName, options => options.JobId = jobId)
+++     .WithAutoRestartOnRotatedCredentials(...)
        .WithServiceBusMessageHandler<OrdersAzureServiceBusMessageHandler, Order>();
--- services.WithServiceBusMessageHandler<OrdersAzureServiceBusMessageHandler, Order>(); <-- No longer possible on services directly
--- services.WithAutoRestartServiceBusQueueMessagePumpOnRotatedCredentials(...); <-- No longer possible on services directly
```

What I would not do, however, is have a `Build` method at the end.

Nothing specifically, but open to suggestions.

**Additional context**

stijnmoreels:
Uses a dedicated type for the Azure Service Bus message handlers and the general message handlers so that the extensions are more discoverable.

Closes #152


stijnmoreels:
Let's see if the separate key rotation is still happy with this, we can rename the `IServiceCollection...` types then.

tomkerkhove (approved):

stijnmoreels:
> Let's see if the separate key rotation is still happy with this, we can rename the `IServiceCollection...` types then.

Gonna change some class names to match better the type their extending.

PR #173, opened 2021-04-15
feat: provide message handling for Azure Functions

stijnmoreels:


stijnmoreels:
@tomkerkhove , some help may be required. I added an Azure Functions project to the solution that runs in a Docker container and uses the message routing/handling to publish the incoming Azure Service Bus message to Azure Event Grid (just like our other Docker test projects).
I get this to work locally, but remotely it seems like the project is not even running, while I (think I) configured this correctly in the YAML definition.

stijnmoreels:
It can be useful to release this version (with the Azure Functions support) before we migrate towards the new Azure SDK bc this new SDK isn't yet supported (as far as I can tell) within Azure Functions with Service Bus trigger.
So, if we release before that, consumers can already use our 'split' of message routing/handling in Azure Functions without to wait for this new SDK to be supported.

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM mcr.microsoft.com/azure-functions/dotnet:3.0 AS base
+WORKDIR /app
+
+FROM mcr.microsoft.com/dotnet/core/runtime-deps:2.2-alpine3.8 AS build
```
tomkerkhove, line 4: Are you using 2.2 on purpose here?

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+WORKDIR "/src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue"
+RUN dotnet build "Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue.csproj" -c Release -o /app/build
+
+FROM build AS publish
+RUN dotnet publish "Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue.csproj" -c Release -o /app/publish
+
+FROM base AS final
+WORKDIR /app
+COPY --from=publish /app/publish .
+CMD [ "/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue" ]
```
tomkerkhove, line 18: This is not correct, see https://github.com/tomkerkhove/azure-friday-keda/blob/main/src/AzureFriday.Shipments/Dockerfile#L10-L15 for and example.

I think you can reproduce your issue by building it locally in your commandline (no VS magic)

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Dockerfile
stijnmoreels: Yeah, I was searching for the right supported image or so.

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Dockerfile
stijnmoreels: Let me check...

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Dockerfile
stijnmoreels: Removed this, resolving.

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Dockerfile
stijnmoreels: Ok, now I'm back at the start, where I got the project running, but it ignores the Azure Function:
https://dev.azure.com/codit/Arcus/_build/results?buildId=52849&view=logs&j=af2770d6-9849-5611-d90b-14595ba0772b&t=b20c33dd-af2d-5d04-7049-688a0f4bc19d

Seems like it's running something else?

tomkerkhove: Did you try to run it outside of VS as a container through the command line?

tomkerkhove: Your issue clearly is the dockerfile. You should use azure function CLI then to generate one

`func init --docker-only` I think

stijnmoreels: Reverted my changes to the default file, which was provided by VS, since the default CLI contains not the necessary copy path commands and I don't know them 😄 .

tomkerkhove: Why do you need the COPY parts?

stijnmoreels: That was defaultly added. I think it's better that you take this Docker-part over from me, as I don't really know what or how this is supposed to be working.

stijnmoreels:
@tomkerkhove , could you help me out or take over this one? As the Docker tests keeps failing for me. It works locally but don't know exactly how my Dockerfile is wrong.
The implementation and stuff is ready for review, only this Docker tests doesn't seem to work. It seems that the test project hosting the Azure Function with message handling is not running.

build/templates/build-and-run-az-func-container.yml
```diff
@@ -0,0 +1,62 @@
+parameters:
+  containerName: ''
+  projectName: ''
+  imageName: ''
+  imageTag: ''
+  port: ''
```
tomkerkhove, line 6: The new parameter format can help you with validation - https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#parameters

But that might be a dedicated PR

src/Arcus.Messaging.AzureFunctions.ServiceBus/Arcus.Messaging.AzureFunctions.ServiceBus.csproj
```diff
@@ -0,0 +1,33 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <LangVersion>latest</LangVersion>
+    <PackageId>Arcus.Messaging.AzureFunctions.ServiceBus</PackageId>
+    <Authors>Arcus</Authors>
+    <Company>Arcus</Company>
+    <Product>Arcus.Messaging</Product>
+    <Description>Provides Azure Service Bus message handling/routing for Azure Functions triggers.</Description>
```
tomkerkhove, line 10:
```suggestion
    <Description>Provides Azure Service Bus message handling/routing for Azure Functions.</Description>
```

src/Arcus.Messaging.AzureFunctions.ServiceBus/Arcus.Messaging.AzureFunctions.ServiceBus.csproj
```diff
@@ -0,0 +1,33 @@
...
+    <PackageProjectUrl>https://github.com/arcus-azure/arcus.messaging</PackageProjectUrl>
+    <PackageIconUrl>https://raw.githubusercontent.com/arcus-azure/arcus/master/media/arcus.png</PackageIconUrl>
+    <RepositoryUrl>https://github.com/arcus-azure/arcus.messaging</RepositoryUrl>
+    <RepositoryType>Git</RepositoryType>
+    <PackageTags>Azure;Messaging;ServiceBus;AzureFunctions</PackageTags>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+
+
```
tomkerkhove, line 22: Nit
```suggestion

```

src/Arcus.Messaging.AzureFunctions.ServiceBus/Arcus.Messaging.AzureFunctions.ServiceBus.csproj
```diff
@@ -0,0 +1,33 @@
...
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
```
tomkerkhove, line 24: Suggestion - Should we use a version range here as well? We can open a new issue to update all the rest if we agree, but if we introduce a new package we can already take this along.

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS installer-env
```
tomkerkhove: See https://github.com/arcus-azure/arcus.messaging/commit/fddac09140aac82d1efd2190268dbcdcd469d341

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/ServiceBusFunction.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        [FunctionName("order")]
+        public async Task Run(
+            [ServiceBusTrigger("docker-az-func-queue", Connection = "ServiceBusConnection")] ServiceBusReceivedMessage message,
+            ILogger log,
+            CancellationToken cancellationToken)
+        {
+            log.LogInformation($"C# ServiceBus queue trigger function processed message: {message.MessageId}");
+            
+            var context = new AzureServiceBusMessageContext(message.MessageId, _jobId, AzureServiceBusSystemProperties.CreateFrom(message), message.ApplicationProperties);
```
tomkerkhove, line 36: What if we provide this?
```suggestion
            var context = AzureServiceBusMessageContext.Create(message, _jobId);
```

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Startup.cs
```diff
@@ -0,0 +1,42 @@
...
+                var eventGridKey = configuration.GetValue<string>("Values:EventGridAuthKey");
+
+                return EventGridPublisherBuilder
+                    .ForTopic(eventGridTopic)
+                    .UsingAuthenticationKey(eventGridKey)
+                    .Build();
+            });
+            
+            builder.AddServiceBusMessageRouting()
+                   .WithServiceBusMessageHandler<OrdersAzureServiceBusMessageHandler, Order>();
```
tomkerkhove, line 40: Let's make sure we track something to have this in our templates as well

src/docker-compose.override.yml
```diff
@@ -17,3 +17,6 @@ services:
       - EVENTGRID_AUTH_KEY=<Event Grid Auth Key>
     ports:
       - '42064:5001'
+  arcus.messaging.tests.runtimes.azurefunction.servicebus.queue:
```
tomkerkhove: We don't need this since Docker Compose doesn't like Azure Functions

https://github.com/microsoft/DockerTools/issues/124

src/docker-compose.yml
```diff
@@ -11,3 +11,8 @@ services:
     build:
       context: .
       dockerfile: Arcus.Messaging.Tests.Workers.ServiceBus.Topic/Dockerfile
+  arcus.messaging.tests.runtimes.azurefunction.servicebus.queue:
```
tomkerkhove: We don't need this since Docker Compose doesn't like Azure Functions

https://github.com/microsoft/DockerTools/issues/124

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -223,35 +287,40 @@ public async Task<MessageResult> TryCustomDeserializeMessageAsync(string message
...
-                    }
+                    _logger.LogTrace("No {MessageBodySerializerType} was found on the registered message handler, so no custom deserialization is available", nameof(IMessageBodySerializer));
+                    return MessageResult.Failure("No custom deserialization was found on the registered message handler");
+                }
 
+                Task<MessageResult> deserializeMessageAsync = serializer.DeserializeMessageAsync(message);
+                if (deserializeMessageAsync is null)
+                {
                     _logger.LogTrace("Invalid {MessageBodySerializerType} message deserialization was configured on the registered message handler, custom deserialization returned 'null'", nameof(IMessageBodySerializer));
                     return MessageResult.Failure("Invalid custom deserialization was configured on the registered message handler");
```
tomkerkhove, line 300: What if I don't use custom deserialization and it returns null, will that also return this message?

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -223,35 +287,40 @@ public async Task<MessageResult> TryCustomDeserializeMessageAsync(string message
...
                 }
+
+                MessageResult result = await deserializeMessageAsync;
+                if (result is null)
+                {
+                    _logger.LogTrace("No {MessageBodySerializerType} was found on the registered message handler, so no custom deserialization is available", nameof(IMessageBodySerializer));
+                    return MessageResult.Failure("No custom deserialization was found on the registered message handler");
```
tomkerkhove, line 307: What if I don't use custom deserialization and it couldn't find a match, will that also return this message?

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -223,35 +287,40 @@ public async Task<MessageResult> TryCustomDeserializeMessageAsync(string message
...
+                if (result.IsSuccess)
+                {
+                    Type deserializedMessageType = result.DeserializedMessage.GetType();
+                    if (deserializedMessageType == MessageType || deserializedMessageType.IsSubclassOf(MessageType))
+                    {
+                        return result;
+                    }
+
+                    _logger.LogTrace("Incoming message '{DeserializedMessageType}' was successfully custom deserialized but can't be processed by message handler because the handler expects message type '{MessageHandlerMessageType}'; fallback to default deserialization", deserializedMessageType.Name, MessageType.Name);
+                    return MessageResult.Failure("Custom message deserialization failed because it didn't match the expected message handler's message type");
```
tomkerkhove, line 319: Same here, what if I don't use custom message deserialization?

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -223,35 +287,40 @@ public async Task<MessageResult> TryCustomDeserializeMessageAsync(string message
...
+                }
+
+                _logger.LogTrace("Custom {MessageBodySerializerType} message deserialization failed: {ErrorMessage} {Exception}", nameof(IMessageBodySerializer), result.ErrorMessage, result.Exception);
+                return MessageResult.Failure("Custom message deserialization failed due to an exception");
```
tomkerkhove, line 323: Same here, what if I don't use custom message deserialization?

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -223,35 +287,40 @@ public async Task<MessageResult> TryCustomDeserializeMessageAsync(string message
...
+                    if (deserializedMessageType == MessageType || deserializedMessageType.IsSubclassOf(MessageType))
+                    {
+                        return result;
+                    }
+
+                    _logger.LogTrace("Incoming message '{DeserializedMessageType}' was successfully custom deserialized but can't be processed by message handler because the handler expects message type '{MessageHandlerMessageType}'; fallback to default deserialization", deserializedMessageType.Name, MessageType.Name);
+                    return MessageResult.Failure("Custom message deserialization failed because it didn't match the expected message handler's message type");
+                }
+
+                _logger.LogTrace("Custom {MessageBodySerializerType} message deserialization failed: {ErrorMessage} {Exception}", nameof(IMessageBodySerializer), result.ErrorMessage, result.Exception);
```
tomkerkhove, line 322: If it's an exception we might want to track it as an exception, no? Would that be because of the code that we wrote?

tomkerkhove (approved):

build/templates/build-and-run-az-func-container.yml
stijnmoreels: Added one here: https://github.com/arcus-azure/arcus/issues/141

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Startup.cs
stijnmoreels: Already here: https://github.com/arcus-azure/arcus.templates/issues/244, resolving.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: This parent method `TryCustomDeserializeMessageAsync` will be called in the router, if it returns a failure result, it will try to do regular JSON deserialization.
Custom deserialization registrations will be `MessageHandlerRegistration` types instead of the `IMessageHandler` instance directly.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: If you don't use custom deserialization, (and no other filters) it won't be registered as an `MessageHandlerRegistration` instance and otherwise, the `Serializer` property will not be filled out. 
The fallback is then general JSON deserialization.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: General fallback to JSON deserialization.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: General fallback to JSON deserialization in the router.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: Ah, yes, that would be better. Thanks.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
tomkerkhove: So it will not hit this point then?

tomkerkhove: If so, perfect

stijnmoreels: > So it will not hit this point then?

Exactly!

PR #193, opened 2021-06-18
feat: add messaging option for more flexible json parsing

### LINKED ISSUE #192: Provide more flexible message parsing in message pump
When adding new properties to the messaging model we do not always remember to update message consumers' projects

Because of that, the messages are no longer being processed by the defined message handler because it shows the following verbose level log:

```
Message handler 'X' is not able to process the message because the incoming message cannot be deserialized to the message that the message handler can handle
```

As a result, the message is moved on and handled by the fallback message handler.

To avoid causing disrupting behavior it would be nice to have a flag that toggles LAX parsing, for example in the extension method: **AddServiceBusQueueMessagePump** adding a parameter i.e.: laxParsingEnabled = false so that existing implementations are not directly impacted but that we could easily switch on.  

It would also be nice to see the above message being presented as an Error in the logs, making it easier to intercept it without having to toggle Logging levels.

**Additional context**
Our message contract: 
```json
{
  "Imei": "10.34.0.16:4096",
  "FirmwareVersion": "efefad25-0d9f-4745-88d9-99b040b80c54",
  "DeviceId": "352554100004969",
  "Ip": "3.32.16.5",
  "MessageReceivedDateTime": "2021-06-15T14:02:35.190Z"
}
```

Received message payload:
```diff
{
  "Imei": "10.34.0.16:4096",
  "FirmwareVersion": "efefad25-0d9f-4745-88d9-99b040b80c54",
  "DeviceId": "352554100004969",
  "Ip": "3.32.16.5",
+ "Status": null,
  "MessageReceivedDateTime": "2021-06-15T14:02:35.190Z"
}
```

stijnmoreels:
Add an message router option to toggle whether or not the JSON deserialization should ignore missing members upon matching incoming messages with registered message handlers.

Closes #192

### DIFF
--- a/docs/preview/features/message-pumps/service-bus.md
+++ b/docs/preview/features/message-pumps/service-bus.md
@@ -185,9 +185,9 @@ public class Startup
                 // (default: Transaction-Id).
                 options.Correlation.TransactionIdPropertyName = "X-Transaction-ID";
 
-                // Indicate whether or not the default built-in JSON deserialization should ignore missing members 
-                // when deserializing the incoming message (default: false).
-                options.Deserialization.IgnoreMissingMembers = true;
+                // Indicate whether or not the default built-in JSON deserialization should ignore additional members 
+                // when deserializing the incoming message (default: AdditionalMemberHandling.Error).
+                options.Deserialization.AdditionalMembers = AdditionalMembersHandling.Ignore;
 
                 // Indicate whether or not a new Azure Service Bus Topic subscription should be created/deleted
                 // when the message pump starts/stops (default: CreateOnStart & DeleteOnStop).
@@ -217,9 +217,9 @@ public class Startup
                 // (default: Transaction-Id).
                 options.Correlation.TransactionIdPropertyName = "X-Transaction-ID";
 
-                // Indicate whether or not the default built-in JSON deserialization should ignore missing members 
-                // when deserializing the incoming message (default: false).
-                options.Deserialization.IgnoreMissingMembers = true;
+                // Indicate whether or not the default built-in JSON deserialization should ignore additional members 
+                // when deserializing the incoming message (default: AdditionalMemberHandling.Error).
+                options.Deserialization.AdditionalMembers = AdditionalMembersHandling.Ignore;
             });
 
         // Uses managed identity to authenticate with the Service Bus Topic:
--- a/docs/preview/features/message-pumps/service-bus.md
+++ b/docs/preview/features/message-pumps/service-bus.md
@@ -185,6 +185,10 @@ public class Startup
                 // (default: Transaction-Id).
                 options.Correlation.TransactionIdPropertyName = "X-Transaction-ID";
 
+                // Indicate whether or not the default built-in JSON deserialization should ignore missing members 
+                // when deserializing the incoming message (default: false).
+                options.Deserialization.IgnoreMissingMembers = true;
+
                 // Indicate whether or not a new Azure Service Bus Topic subscription should be created/deleted
                 // when the message pump starts/stops (default: CreateOnStart & DeleteOnStop).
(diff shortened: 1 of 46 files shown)


docs/preview/features/message-pumps/service-bus.md
```diff
@@ -185,6 +185,10 @@ public class Startup
                 // (default: Transaction-Id).
                 options.Correlation.TransactionIdPropertyName = "X-Transaction-ID";
 
+                // Indicate whether or not the default built-in JSON deserialization should ignore missing members 
+                // when deserializing the incoming message (default: false).
+                options.Deserialization.IgnoreMissingMembers = true;
```
tomkerkhove, line 190: I think it's a different perspective but what about this?
```suggestion
                options.Deserialization.IgnoreAdditionalFields = true;
```

This feels a bit more clear that it's aiming towards additional info in the payload. What do you think @fgheysels ?

src/Arcus.Messaging.Abstractions/MessageHandling/MessageDeserializationOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// <summary>
+    /// Represents the consumer-configurable options, as part of the <see cref="MessageRouterOptions"/>, to change the behavior of the <see cref="MessageRouter"/>.
+    /// </summary>
+    public class MessageDeserializationOptions
+    {
+        /// <summary>
+        /// Gets or sets the flag indicating whether or not the default JSON deserialization should ignore missing members
+        /// when trying to match a incoming message with a message handler's message type.
+        /// </summary>
+        public bool IgnoreMissingMembers { get; set; } = false;
```
tomkerkhove, line 12: Same suggestion here

tomkerkhove (approved):

docs/preview/features/message-pumps/service-bus.md
stijnmoreels: We could create some email for this too. I created this one but seems like an enum is maybe better as other things could/should be configurable for consumers.

tomkerkhove: What do you mean with email? Or did you mean enum?

Enum is fine for me as well.

stijnmoreels: Yeah, enum 😅

src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouter.cs
```diff
@@ -334,5 +381,27 @@ protected virtual bool TryDeserializeToMessageFormat(string message, Type messag
...
+                    Logger.LogTrace("JSON deserialization uses 'Ignore' result when encountering additional members");
+                    jsonSerializer.MissingMemberHandling = MissingMemberHandling.Ignore;
+                    break;
+                default:
+                    Logger.LogTrace("JSON deserialization uses 'Error' result when encountering additional members");
+                    jsonSerializer.MissingMemberHandling = MissingMemberHandling.Error;
+                    break;
+            }
+
+            return jsonSerializer;
```
tomkerkhove, line 404:
```suggestion
            switch (Options.Deserialization?.AdditionalMembers)
            {
                case AdditionalMemberHandling.Error:
                    jsonSerializer.MissingMemberHandling = MissingMemberHandling.Error;
                    break;
                case AdditionalMemberHandling.Ignore:
                    jsonSerializer.MissingMemberHandling = MissingMemberHandling.Ignore;
                    break;
                default:
                    jsonSerializer.MissingMemberHandling = MissingMemberHandling.Error;
                    break;
            }
            
            Logger.LogTrace($"JSON deserialization uses '{jsonSerializer.MissingMemberHandling}' result when encountering additional members");

            return jsonSerializer;
```

PR #196, opened 2021-06-23
feat: use message handler options to pass-in az servicebus receiver

### LINKED ISSUE #195: Move `ServiceBusReceiver` to options model for furture-proof message routing
Move our `ServiceBusReceiver` model from the router signature to an options model so that we are more safe in the future when we want to add stuff from the Azure Functions/message pump to the router.

Adding new stuff to the signature, but that requires breaking changes.

stijnmoreels:
Pass-in the `ServiceBusReceiver` via an options model, so can more easily change things later.

Closes #195

### DIFF
--- a/src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageHandlerOptions.cs
+++ b/src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageHandlerOptions.cs
@@ -1,0 +1,16 @@
+﻿using Azure.Messaging.ServiceBus;
+
+namespace Arcus.Messaging.Abstractions.ServiceBus.MessageHandling
+{
+    /// <summary>
+    /// Represents the additional options to change the behavior of the Azure Service Bus message handling during the routing of the <see cref="AzureServiceBusMessageRouter"/>.
+    /// </summary>
+    public class AzureServiceBusMessageHandlerOptions
+    {
+        /// <summary>
+        /// Gets or sets the instance that can call operations (dead letter, complete...) on an Azure Service Bus <see cref="ServiceBusReceivedMessage"/>;
+        /// used within <see cref="AzureServiceBusMessageHandler{TMessage}"/>s or <see cref="AzureServiceBusFallbackMessageHandler"/>s with Azure Service Bus specific operations.
+        /// </summary>
+        public ServiceBusReceiver ServiceBusReceiver { get; set; }
+    }
+}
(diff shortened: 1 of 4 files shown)


stijnmoreels:
Let me know what you think of this solution. Consumers that uses our router outside the message pump, and uses specific Azure Service Bus operations (dead letter, complete, ...) will have to pass-in the `ServcieBusReceiver` via the options now.

tomkerkhove:
As and end-user, I won't see this change - Right?

stijnmoreels:
> As and end-user, I won't see this change - Right?

When using the message pump, yeah , no chance there. If we support Azure Functions in the future, they will have to pass it along via the options.

tomkerkhove:
> When using the message pump, yeah , no chance there. If we support Azure Functions in the future, they will have to pass the receiver along via the options.
How is that? Wouldn't we create it for them? Because in a function there is no such thing as the client but rather done through the bindings, no?

stijnmoreels:
> When using the message pump, yeah , no chance there. If we support Azure Functions in the future, they will have to pass the receiver along via the options.
> > How is that? Wouldn't we create it for them? Because in a function there is no such thing as the client but rather done through the bindings, no?

Right now, if we want to make use of our message handlers in an Azure Function, it would look like this:

```csharp
public class OrderFunction
{
    private readonly IAzureServiceBusMessageRouter _messageRouter;

    public OrderFunction(IAzureServiceBusMessageRouter messageRouter)
    {
        _messageRouter = messageRouter;
        JobId = Guid.NewGuid().ToString();
    }

    public string JobId { get; }

    [FunctionName("orders")]
    public async Task Run(
        [ServiceBusTrigger("queue-name", Connection = "ConnectionString")] ServiceBusReceivedMessage message,
        ServiceBusReceiver receiver,
        ILogger logger,
        CancellationToken cancellationToken)
    {
        var properties = AzureServiceBusSystemProperties.CreateFrom(message);
        var context = new AzureServiceBusMessageContext(message.MessageId, JobId, properties, message.ApplicationProperties); 
        MessageCorrelationInfo correlationInfo = message.GetCorrelationInfo();

        // 1. routing message without the possibility to call 'dead letter', 'complete' during the message handling.
        await _messageRouter.RouteMessageAsync(message, context, correlationInfo, cancellationToken);

        // 2. routing message with the possibilty to call 'dead letter', 'complete' during the message handling.
        await _messageRouter.RouteMessageAsync(message, context, correlationInfo, cancellationToken, options =>
        {
            options.ServiceBusReceiver = receiver;
        });
    }
}
````

tomkerkhove:
I'd love to make 2 more user friendly though. Options is good for internal stuff but for this I'd rather have it as a parameter since it's "required".

stijnmoreels:
> I'd love to make 2 more user friendly though. Options is good for internal stuff but for this I'd rather have it as a parameter since it's "required".

Ok, then we can actually abandon this PR, as that's how it now is.

PR #214, opened 2021-08-12
feat: make message routing options public for custom routers

### LINKED ISSUE #212: Provide access to the message router options via the message pump options for custom message router implementations
When implementing a custom `IAzureServiceBusMessageRouter`, we can't use the same `configureMessagePump` function (message router options are part of the message pump options in this case), because the message router options are `internal` in the message pump options.

We should consider making this `MessageRouterOptions` property either `public` or something in this line, so we can re-use the `configureMessagePump` function and therefore configure the message router the same way.

**Additional context**
https://github.com/arcus-azure/arcus.messaging/blob/7b46969afb2638563152878203e10bd1099d28fb/src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptions.cs#L112

stijnmoreels:
When creating custom message router (like the CloudEvents message router in the background jobs repo), we could benefit from the message pump options and therefore the message router options, but then we need to have access to the message router options directly.

Closes #212

### DIFF
--- a/src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptions.cs
+++ b/src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptions.cs
@@ -104,12 +104,13 @@ namespace Arcus.Messaging.Pumps.ServiceBus.Configuration
         /// <summary>
         /// Gets the consumer-configurable options to change the deserialization behavior.
         /// </summary>
-        public MessageDeserializationOptions Deserialization => MessageRouterOptions.Deserialization;
+        [Obsolete("Use the " + nameof(Routing) + ".Deserialization instead")]
+        public MessageDeserializationOptions Deserialization => Routing.Deserialization;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -104,12 +104,13 @@ public int MaximumUnauthorizedExceptionsBeforeRestart
         /// <summary>
         /// Gets the consumer-configurable options to change the deserialization behavior.
         /// </summary>
-        public MessageDeserializationOptions Deserialization => MessageRouterOptions.Deserialization;
+        [Obsolete("Use the " + nameof(Routing) + ".Deserialization instead")]
+        public MessageDeserializationOptions Deserialization => Routing.Deserialization;
```
stijnmoreels, line 108: Deprecation message not present when configuring the message pump, as they are using interfaces to provide the correct set of options. This deprecation message is more for contributors and consumers that want to write custom message routers.

tomkerkhove (approved):

PR #215, opened 2021-08-13
fix: Correctly pass information to logs for message processing failure

tomkerkhove:


stijnmoreels (approved):
Woops, that was no good. Thanks for the catch!

PR #219, opened 2021-09-28
fix: Update Newtonsoft.Json package reference to use 12.0.1

tomkerkhove:
Update Newtonsoft.Json package reference to use 12.0.1 since 12.0.0 does not exist

This is causing build warning:
> [...] Warning NU1603: Arcus.Messaging.Abstractions 1.0.1 depends on Newtonsoft.Json (>= 12.0.0) but Newtonsoft.Json 12.0.0 was not found. An approximate best match of Newtonsoft.Json 12.0.1 was resolved.


stijnmoreels:
Yes, was something I mentioned to you earlier. 😄 . Glad that we can fix this.

stijnmoreels (approved):
Wondering if we need package ranges for this...

PR #223, opened 2021-10-19
feat: add dedicated message correlation accessor

stijnmoreels:
Adds an dedicated `IMessageCorrelationInfoAccessor` so we are more certain of the scoped correlation in a combined Arcus-correlation situation.

Relates to https://github.com/arcus-azure/arcus/issues/162


stijnmoreels:
Hmm, beter to verify the service collection too.

stijnmoreels:
I'll add some docs too.

PR #227, opened 2021-11-30
fix: use camelCase for tcp health probe serialization

### LINKED ISSUE #226: Use camelCase for default `HealthReport` serialization in the TCP health probe
When serializing the `HealthReport` upon calling the TCP health probe, we don't use camelCase by default as JSON serialization settings.

As we use by-default camelCase in the templates repo, we should make sure that we streamline consistency and also use camelCase here.

stijnmoreels:
We should use camelCasing for the TCP health probe to be consistent with our other repo's JSON settings (like the templates repo).

Closes #226

### DIFF
--- a/src/Arcus.Messaging.Health/Tcp/TcpHealthListener.cs
+++ b/src/Arcus.Messaging.Health/Tcp/TcpHealthListener.cs
@@ -13,6 +13,7 @@ using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Serialization;
 
 namespace Arcus.Messaging.Health.Tcp
 {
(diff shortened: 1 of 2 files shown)


fgheysels (approved):

fgheysels:
Good to be consistent, but this might be a breaking change ?

stijnmoreels:
> Good to be consistent, but this might be a breaking change ?

Good question. Not sure if it qualifies as a breaking change as a decent JSON serialization would maybe take this into account. Not sure.

fgheysels:
> > Good to be consistent, but this might be a breaking change ?
> 
> Good question. Not sure if it qualifies as a breaking change as a decent JSON serialization would maybe take this into account? Not sure.

Json is case-sensitive.

stijnmoreels:
> > > Good to be consistent, but this might be a breaking change ?
> > 
> > 
> > Good question. Not sure if it qualifies as a breaking change as a decent JSON serialization would maybe take this into account? Not sure.
> 
> Json is case-sensitive.

Yeah, I was talking about looser serialization.
Hmm, maybe we then hold off with this change?

fgheysels:
> > > > Good to be consistent, but this might be a breaking change ?
> > > 
> > > 
> > > Good question. Not sure if it qualifies as a breaking change as a decent JSON serialization would maybe take this into account? Not sure.
> > 
> > 
> > Json is case-sensitive.
> 
> Yeah, I was talking about looser serialization. Hmm, maybe we then hold off with this change?

I think we can have breaking changes, we just need to be open / communicative about them.

stijnmoreels:
Close for now.

Issue #229, opened 2022-01-12
Remove 'auto restart message pump on rotated credentials'


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
We have decided to move the `auto restart message pump on rotated credentials' to the background jobs projects bc it's a better place for it. (Also, background jobs is one of its dependencies). This project should therefore be removed from the library.

**Describe the solution you'd like**
* Mark the available functionality as obsolete.
* Deprecate the NuGet package.

**Additional context**
Functionality is maintained here: https://background-jobs.arcus-azure.net/Features/Security/auto-restart-servicebus-messagepump-on-rotated-credentials

PR #230, opened 2022-01-12
chore: mark keyrotation as obsolete

stijnmoreels:
Mark the current key rotation functionality as obsolete as this is moved to Arcus BackgroundJobs.

Relates to #229


fgheysels (approved):
Good that it is explicitly stated that this is obsolete

PR #260, opened 2022-03-22
chore: update towards net6.0

stijnmoreels:
Add .NET 6 support, whilst still supporting .NET Core.

Relates to https://github.com/arcus-azure/arcus/issues/106


stijnmoreels:
@fgheysels , not sure what I'm doing wrong here: https://dev.azure.com/codit/Arcus/_build/results?buildId=60218&view=logs&j=af2770d6-9849-5611-d90b-14595ba0772b&t=d498e235-96e9-5261-421a-d6ea71f144f9
My Docker knowledge fails me here 😢 .

stijnmoreels:
> > @fgheysels , not sure what I'm doing wrong here: https://dev.azure.com/codit/Arcus/_build/results?buildId=60218&view=logs&j=af2770d6-9849-5611-d90b-14595ba0772b&t=d498e235-96e9-5261-421a-d6ea71f144f9 My Docker knowledge fails me here 😢 . The .NET Core requirement keeps bugging.
> 
> Maybe you need to update the Functions SDK (Microsoft.NET.Sdk.Functions) from 3.0.11 to 4.1.0 ?

🙌 Yes! That was it. Thanks for this! Was not paying attention to that.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -106,19 +106,29 @@ private static Type[] GetServiceRegistrationTypes(IServiceProvider serviceProvid
 
         private static IEnumerable<Type> GetServiceRegistrationTypesFromMicrosoftExtensions(IServiceProvider serviceProvider, ILogger logger)
         {
-            object engine = 
-                serviceProvider.GetType().GetProperty("Engine")?.GetValue(serviceProvider) 
-                ?? serviceProvider.GetType().GetField("_engine", BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(serviceProvider);
+            object engine =
```
stijnmoreels, line 109: You'll probably wonder why we have this reflection code in the `MessageHandler` file.
This is bc we were faced with a problem when implementing the message routing. The `IServceProvider` doesn't support open generic interfaces and since our `IMessageHandler<,>` is such an open interface, there is no build-in way to extract all service registrations with a generic message type.
This message handler class will do this for us. It searches for the 'engine' and 'call site factory' on the service provider and extracts all the message handler interfaces. Later on, we can determine if there exists and message handler that can process the incoming message.

We can off course discuss this later, or come up with something else. For now, this was the best we could do. Unfortunately.

fgheysels: I was indeed wondering,  It might be good to have a short call on this once upon a time to explain this :)

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -129,17 +139,40 @@ private static IEnumerable<Type> GetServiceRegistrationTypesFromMicrosoftExtensi
...
+
+            return callSiteFactory;
+        }
+
         private static IEnumerable<Type> GetServiceRegistrationTypesFromWebJobs(IServiceProvider serviceProvider, ILogger logger)
         {
-            object container = serviceProvider.GetType().GetField("_container", BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(serviceProvider);
+            object containerNetCore = serviceProvider.GetOptionalFieldValue("_container");
+            object containerNet6 = serviceProvider.GetOptionalFieldValue("_resolver");
+            object container = containerNetCore ?? containerNet6;
```
fgheysels, line 167: For me, this is also a bit obscure.  Why the fields _resolver and _container ?  And how are they related with 'containerNetCore' and containerNet6 ?

src/Arcus.Messaging.Tests.Integration/Health/TcpHealthService.cs
```diff
@@ -48,14 +53,45 @@ public async Task<HealthReport> GetHealthReportAsync()
...
+                       IEnumerable<KeyValuePair<string, HealthReportEntry>> reportEntries = entries.Children().Select(CreateHealthReportEntry);
+
+                        var report = new HealthReport(
+                            new ReadOnlyDictionary<string, HealthReportEntry>(reportEntries.ToDictionary(entry => entry.Key, entry => entry.Value)),
+                            Enum.Parse<HealthStatus>(status.Value<string>()),
+                            TimeSpan.Parse(totalDuration.Value<string>()));
+
+                        _logger.LogTrace("Health report retrieved");
+                        return report;
+                    }
```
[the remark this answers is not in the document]
stijnmoreels: Ah, right, I had to change this test service as the constructor changed on the `HealthReport` type and the JSON deserialization failed. I'll clean this up a bit.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: This is because the fields were renamed and that versions. So, in .NET Core it was called `_container` but in .NET 6 they renamed the field to `_resolver`.

stijnmoreels: It's not very stable, I know, it's the best we can do though. We can/should discus this later, for sure.

src/Arcus.Messaging.Tests.Integration/Health/TcpHealthService.cs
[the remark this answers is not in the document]
stijnmoreels: Ah, yeah, I pushed the clean-up just now. Thx for reminding. 👍

fgheysels (approved):
Thanks!

PR #266, opened 2022-04-06
feat: add az servicebus request tracking to message router

### LINKED ISSUE #221: Provide Azure Service Bus request tracking in message pump
We should track the incoming Azure Service Bus request as telemetry in our message pump so the parent operation ID is also linked to this.

Alter the message pump so we can track incoming Azure Service Bus requests.

When using Azure Functions, and only our router, we should maybe consider moving the tracking functionality to the router instead of in the pump.

**Additional context**
Observability will provide us with the right extension: https://github.com/arcus-azure/arcus.observability/issues/242

stijnmoreels:
Adds Azure Service Bus request tracking to the message router so that every processed message is also tracked as an incoming request in Azure Application Insights.

Closes #221


fgheysels (approved):
Small but important change :)

PR #267, opened 2022-04-08
fix: add correlation registration in message router

### LINKED ISSUE #265: Move correlation assignment to the message routing
Since we will be able to support message handling in Azure Functions, there are some missing links with the assignment of the message correlation to the correlation accessor. Currently, this is being done in the message pump but since they're is not pump in an Azure Functions context, this should be moved.

Move the message correlation assignment to the message router so that each new message correlation is picked up by the correlation accessor and is available for the remaining message handlers.

stijnmoreels:
Since we need scoped correlation registration outside the message pump (see Azure Functions, where there is no message pump), we need to move to scoped correlation registration to the message router.

Closes #265


stijnmoreels:
We'll make sure in another PR that we can configure the correlation options too within the router options and not in the pump options. Seems that it's already complex enough, this PR, so we can do it in steps. 😉

fgheysels (approved):

Issue #258, opened 2022-03-17
Rework message pump/message handler docs approach for both Azure Service Bus and general approach


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
We have a very thorough explanation of how the Azure Service Bus message pump works and how people should approach this with their own message handler implementations. We do not have, though, a guide about Azure Service Bus message handlers specific. A problem was raised (#254 ) that was caused due to too-general message handler registration. This was fixed by adding a message context predicate to the message handler registration but this way or registering is only briefly mentioned when talking about message handlers in general (https://messaging.arcus-azure.net/Features/message-pumps/customization). 

**Describe the solution you'd like**
We should rework how we approach message handler/message pumps in both general and Azure Service Bus variants. It's a good thing that we focus on the message handling as this is where the consumers will spent most of their time. We should take a look at the existing headers and splitting and simplify this, keeping in mind that we do not want dead links.

**Additional context**
Follow-up of #254

stijnmoreels:
A PR with the possible redirects is necessary when publishing the new version.

stijnmoreels:
Redirects are already added to [`netlify.toml`](https://github.com/arcus-azure/arcus.messaging/blob/main/netlify.toml), if one navigate to https://messaging.arcus-azure.net/features/message-pumps/service-bus, they will be redirected to https://messaging.arcus-azure.net/features/message-handling/service-bus/.

PR #268, opened 2022-04-15
docs: rework message handling structure

stijnmoreels:
Reworks the message handling feature documentation with a better/cleaner message handling focus points and visuals. Unlike before, we focus now on message handling instead of pumps, so we should reflect this in the feature docs.

These changes in the preview feature docs also contains some page changes, so we will need to add some redirects when we release a new version of this.

Relates to #258


docs/preview/01-index.md
```diff
@@ -17,13 +17,13 @@ PM > Install-Package Arcus.Messaging.Abstractions
...
-- Support for using message pumps for the following brokers:
-    - Azure Service Bus ([docs](./02-Features/message-pumps/service-bus.md) | [extensions](./02-Features/service-bus.md))
-- Support for exposing TCP health probes to periodically check liveness/readiness of the host ([docs](./02-Features/tcp-health-probe.md))
-- Customize message pumps ([docs](./02-Features/message-pumps/customization.md))
-    - Fallback message handlers
-- Using message pumps with Azure Functions
-    - Azure Service Bus ([docs](./02-Features/message-pumps/service-bus-azure-functions.md))
+- Support for using message handling for the following brokers:
+  - Azure Service Bus
+    - Worker ([docs](./02-Features/02-message-handling/01-service-bus.md))
```
fgheysels, line 22: What exactly is meant by 'worker'.  This is actually a background worker that processes messages from the bus.
I wonder if we can come up with a better name.

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -140,111 +131,105 @@ public class Startup
...
+}
+```
 
-        // Specify the name of the Service Bus Topic, and provide a name for the Topic subscription:
-        services.AddServiceBusMessageTopicMessagePump<OrdersMessageHandler>(
-            "My-Service-Bus-Topic-Name",
-            "My-Service-Bus-Topic-Subscription-Name",
-            "ARCUS_SERVICEBUS_ORDERS_CONNECTIONSTRING");
+> Note that the order in which the message handlers are registered is important in the message processing.
+> In the example, when a message handler above this one is registered that could also handle the message (same message type) than that handler may be chosen instead of the one with the specific filter.
```
fgheysels, line 140: may be chosen or will be chosen ?

The first option implies that you're not certain which handler will be used at runtime.  The latter means that you're absolutely sure that the first registered handler will be used.

I'd extend line 139 with an additional sentence that says that the first registered message-handler that is eligeable to process a certain message, will be used.

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -140,111 +131,105 @@ public class Startup
...
-        // Uses managed identity to authenticate with the Service Bus Topic:
-        services.AddServiceBusTopicMessagePumpUsingManagedIdentity(
-            topicName: properties.EntityPath,
-            subscriptionName: "Receive-All", 
-            fullyQualifiedNamespace: "<your-namespace>.servicebus.windows.net"
-            // The optional client id to authenticate for a user assigned managed identity. More information on user assigned managed identities cam be found here:
-            // https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#how-a-user-assigned-managed-identity-works-with-an-azure-vm
-            clientId: "<your-client-id>");
+### Bring your own deserialization
+You can also choose to extend the built-in message deserialization with additional deserializer to meet your needs.
```
fgheysels, line 144:
```suggestion
You can also choose to extend the built-in message deserialization with a custom deserializer to meet your needs. 
```

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -140,111 +131,105 @@ public class Startup
...
 
-        services.AddServiceBusTopicMessagePump(
-            "ARCUS_SERVICEBUS_ORDERS_CONNECTIONSTRING",
-            options => 
-            {
-                // Indicate whether or not messages should be automatically marked as completed 
-                // if no exceptions occurred and processing has finished (default: true).
-                options.AutoComplete = true;
+You start by implementing an `IMessageBodySerializer`. The following example shows how an expected type can be transformed to something else. 
+The result type (in this case `OrderBatch`) will be then be used to check if there is an `IAzureServiceBusMessageHandler` registered with that message type.
```
fgheysels, line 148:
```suggestion
The result type (in this case `OrderBatch`) will then be used to check if there is an `IAzureServiceBusMessageHandler` registered for that message type.
```

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -140,111 +131,105 @@ public class Startup
...
-                options.EmitSecurityEvents = true;
+```csharp
+using Arcus.Messaging.Pumps.Abstractions.MessageHandling;
 
-                // The amount of concurrent calls to process messages 
-                // (default: null, leading to the defaults of the Azure Service Bus SDK message handler options).
-                options.MaxConcurrentCalls = 5;
+public class OrderBatchMessageBodySerializer : IMessageBodySerializer
+{
+    public async Task<MessageResult> DeserializeMessageAsync(string messageBody)
```
fgheysels, line 155: Should the blank line between 154 & 155 be removed ?

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -140,111 +131,105 @@ public class Startup
...
+            return MessageResult.Success(new OrderBatch(orders));
+        }
+    }
+}
+```
 
-                // The unique identifier for this background job to distinguish 
-                // this job instance in a multi-instance deployment (default: guid).
-                options.JobId = Guid.NewGuid().ToString();
+The registration of these message body handlers can be done just as easily as an `IMessageSerializer`:
```
fgheysels, line 167: This sentence is confusing.  We just showed an example of a custom serializer, but this sentence says that registering a body handler (message handler I suppose ? ) is as simple as registering a serializer.

I think you want to say here that registering a serializer is as easy as registering a message-handler ?

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -140,111 +131,105 @@ public class Startup
...
-                // Indicate whether or not messages should be automatically marked as completed 
-                // if no exceptions occurred and processing has finished (default: true).
-                options.AutoComplete = true;
+> Note that the order in which the message handlers are registered is important in the message processing.
+> In the example, when a message handler above this one is registered that could also handle the message (same message type) than that handler may be chosen instead of the one with the specific filter.
 
-                // Indicate whether or not the message pump should emit security events (default: false).
-                options.EmitSecurityEvents = true;
+### Filter messages based on message body
+When registering a new message handler, one can opt-in to add a filter on the incoming message body which filters out messages that are not needed to be processed.
```
fgheysels, line 193: Can those messages that are filtered out, still be processed by another message handler ?

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -398,8 +380,124 @@ public class Startup
 }
 ```
 
-## Alternative Service Bus message routing
+## Pump Configuration
+Next to that, we provide a **variety of overloads** to allow you to:
+- Specify the name of the queue/topic
+- Only provide a prefix for the topic subscription, so each topic message pump is handling messages on separate subscriptions
+- Configure how the message pump should work *(ie. max concurrent calls & auto delete)*
+- Read the connection string from the configuration *(although we highly recommend using the [Arcus secret store](https://security.arcus-azure.net/features/secret-store) instead)*
```
fgheysels, line 388: Maybe [this](https://docs.microsoft.com/en-us/azure/app-service/app-service-key-vault-references#reference-syntax) syntax is supported.

docs/preview/01-index.md
stijnmoreels: Yeah, like we did in templates, we make a distinction between .NET SDK Workers and Azure Functions. So, seemed that we should keep this distinction here, too.

stijnmoreels:
Ok, so all comments are on the existing docs and not on the restructure in this PR, 😄 . Ok, but I think we can streamline your suggestions on the other versions as well.

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: > may be chosen or will be chosen ?

> The first option implies that you're not certain which handler will be used at runtime. The latter means that you're absolutely sure that the first registered handler will be used.

It's still 'may' here, I think, as the message type is not solely responsible for the message routing. One can add additional message type filters, message body filters, message context filters and even bring in their own serialization for the same message type.

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Not sure what you're referring to. The line here doesn't contain a blank and the code sample doesn't have a blank besides between the usings and the class.

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Ah, yes, 'as easily as a message handler' is correct. Thx.

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Yes, of course, this is just for message routing, not for the actual processing.

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -140,111 +131,105 @@ public class Startup
...
+    {
+        services.WithServiceMessageHandler<OrderMessageHandler, Order>((Order order) => order.Type == Department.Sales);
     }
 }
 ```
 
-## Fallback message handling
-
+### Fallback message handling
 When receiving a message on the message pump and none of the registered `IAzureServiceBusMessageHandler`'s can correctly process the message, the message pump normally throws and logs an exception.
```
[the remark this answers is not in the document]
stijnmoreels: No, not necessary. If the fallback message handler is configured for dead-lettering or abandoning, with our Azure Service Bus template classes, then yes, but this is not done for you. Also, the fallback message handlers are more than only this, and can be used outside Azure Service Bus contexts and can contain more logic then only dead-lettering.

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -398,8 +380,124 @@ public class Startup
 }
 ```
 
-## Alternative Service Bus message routing
+## Pump Configuration
+Next to that, we provide a **variety of overloads** to allow you to:
+- Specify the name of the queue/topic
+- Only provide a prefix for the topic subscription, so each topic message pump is handling messages on separate subscriptions
```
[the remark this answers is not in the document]
stijnmoreels: This refers to the different ways a message pump can be registered in the application. One can choose to use a queue or topic subscription. When choosing a topic subscription, one can also opt to only pass-in a prefix that the topic subscription subscription will have, the rest will be generated.

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: How would this help? This is about the Arcus secret store, right?

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/268/commits/07b70126c45b9737cd2dd37e77f12ee51b1a5e5f

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/268/commits/cd44eb2753ac33d70bef057cea45e54bbf30fb40

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/268/commits/cd44eb2753ac33d70bef057cea45e54bbf30fb40

docs/preview/01-index.md
fgheysels: I agree that there must be a distinction, but I was just thinking if we could come up with a better name then just 'worker'.  (Background Worker, i dunno)

docs/preview/02-Features/02-message-handling/01-service-bus.md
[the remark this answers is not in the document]
fgheysels: Is the topic-subscription then created by the messagepump as well ?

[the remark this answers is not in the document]
stijnmoreels: Yes, it is. 👍

stijnmoreels:
> > Ok, so all comments are on the existing docs and not on the restructure in this PR, 😄 . Ok, but I think we can streamline your suggestions on the other versions as well.
> 
> What do you mean ? This PR only involves docs.

Yes, but it has a whole new way of structuring. Different paths, different places. Some sections together to focus on message handling instead of pumps. Some message handling docs that were previously only mentioned for general message handlers are now also available for Azure Service Bus message handlers. Just saw that all the comments were nothing to do with these changes, but with the existing docs from the previous versions.

docs/preview/01-index.md
stijnmoreels: Maybe 'Worker Service`, then? Like Microsoft describes them: https://docs.microsoft.com/en-us/dotnet/core/extensions/workers

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Yeah, probably. I don't see this, but I got my browser zoom always on 75% 😅 . Formatting is best looked at in the deploy preview instead, since there's the final result.

docs/preview/02-Features/02-message-handling/01-service-bus.md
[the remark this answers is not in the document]
stijnmoreels: If no message handler is able to process it, and no general or Azure Service Bus-specific fallback message handler is configured, and the message pump sets the message on 'auto-complete', then the message will be send back to the queue, that's right. But, this setup would be considered invalid as the message pump will then try again to process the message. Always having a fallback message handler is a good approach.

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Ah, that's cool, but maybe we should still refer to our secret store to make the difference clear between configuration data <> secrets. Could be further investigated, of course.

docs/preview/01-index.md
fgheysels: Indeed, that would be great.

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Great!

docs/preview/01-index.md
stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/268/commits/6d546c732af7afb94c39e8d3e190773988961da5

docs/preview/02-Features/02-message-handling/01-service-bus.md
stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/268/commits/8c5291057252421fb3e696bc84e314c7e9187def

docs/preview/02-Features/02-message-handling/01-service-bus.md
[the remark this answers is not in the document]
fgheysels: Ah, so you can make sure that the code does not create the subscription.  (I am not a fan of that :D )

fgheysels:
> > > Ok, so all comments are on the existing docs and not on the restructure in this PR, 😄 . Ok, but I think we can streamline your suggestions on the other versions as well.
> > 
> > 
> > What do you mean ? This PR only involves docs.
> 
> Yes, but it has a whole new way of structuring. Different paths, different places. Some sections together to focus on message handling instead of pumps. Some message handling docs that were previously only mentioned for general message handlers are now also available for Azure Service Bus message handlers. Just saw that all the comments were nothing to do with these changes, but with the existing docs from the previous versions.
> 
> Maybe with the preview: https://deploy-preview-268--arcus-messaging.netlify.app/next/ you'lll more see what I'm talking about.

I think the new structure is OK.
There's just something that might be missing imho:  a short introduction that describes what this Messaging library can be used for.   Something for in a separate PR.

fgheysels (approved):

fgheysels:
> > > > Ok, so all comments are on the existing docs and not on the restructure in this PR, 😄 . Ok, but I think we can streamline your suggestions on the other versions as well.
> > > 
> > > 
> > > What do you mean ? This PR only involves docs.
> > 
> > 
> > Yes, but it has a whole new way of structuring. Different paths, different places. Some sections together to focus on message handling instead of pumps. Some message handling docs that were previously only mentioned for general message handlers are now also available for Azure Service Bus message handlers. Just saw that all the comments were nothing to do with these changes, but with the existing docs from the previous versions.
> > Maybe with the preview: https://deploy-preview-268--arcus-messaging.netlify.app/next/ you'lll more see what I'm talking about.
> 
> I think the new structure is OK. There's just something that might be missing imho: a short introduction that describes what this Messaging library can be used for. Something for in a separate PR.

Created an issue for this: https://github.com/arcus-azure/arcus.messaging/issues/278

stijnmoreels:
> Created an issue for this: #278

Ok, great! 👍 Will help indeed.

PR #273, opened 2022-04-26
feat: extract operation parent ID from received message

### LINKED ISSUE #269: Extend the `.GetCorrelationInfo` Service Bus message extension to allow operation parent ID extraction
Currently, we are able to generate a `MessageCorelationInfo` model from a received Azure Service Bus message with an operation ID and transaction ID. But we aren't able to extract the operation parent ID from the `UserProperties` like the others.

Provide a way to extract the operation parent ID from the received Azure Service Bus message with as default user property name: `Operation-Parent-Id`.

**Additional context**
Related to PoC: https://github.com/arcus-azure/arcus-service-to-service-correlation-poc

stijnmoreels:
Fully correlation options and operation parent ID extraction in both worker and Azure Functions environment. Updated message creation with simpler correlation settings.

Closes #269


src/Arcus.Messaging.ServiceBus.Core/Extensions/MessageExtensions.cs
```diff
@@ -76,10 +76,41 @@ public static MessageCorrelationInfo GetCorrelationInfo(this ServiceBusReceivedM
...
+            Guard.NotNull(message, nameof(message), "Requires an received Azure Service Bus message to retrieve the correlation information");
+            Guard.NotNullOrWhitespace(transactionIdPropertyName, nameof(transactionIdPropertyName), "Requires a non-blank property name to retrieve the correlation transaction ID from the received Azure Service Bus message");
+            Guard.NotNullOrWhitespace(operationParentIdPropertyName, nameof(operationParentIdPropertyName), "Requires a non-blank property name to retrieve the correlation operation parent ID from the received Azure Service Bus message");
+
             string transactionId = DetermineTransactionId(message, transactionIdPropertyName);
             string operationId = DetermineOperationId(message.CorrelationId);
+            string operationParentId = message.GetOptionalUserProperty(operationParentIdPropertyName);
 
-            var messageCorrelationInfo = new MessageCorrelationInfo(operationId, transactionId);
+            var messageCorrelationInfo = new MessageCorrelationInfo(operationId, transactionId, operationParentId);
```
stijnmoreels, line 113: Most important change here.

src/Arcus.Messaging.Abstractions/MessageCorrelationInfo.cs
```diff
@@ -4,31 +4,41 @@
...
             CycleId = Guid.NewGuid().ToString("D");
         }
+
+        /// <summary>
+        /// Gets the unique identifier that indicates an attempt to process a given message.
+        /// </summary>
+        /// <remarks>
+        ///     If the same message is processed n-times, it will have the same OperationId but n different cycle ids
+        /// </remarks>
+        public string CycleId { get; }
```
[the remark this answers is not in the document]
stijnmoreels: It would be a breaking change, though. It's still more an ID than a counter as it's more linked to the message pump/processing than the amount of times it's processed. So, it's more about the 'cycle', then, maybe.

src/Arcus.Messaging.ServiceBus.Core/ServiceBusMessageBuilder.cs
```diff
@@ -0,0 +1,156 @@
...
+                {
+                    { PropertyNames.ContentType, "application/json" },
+                    { PropertyNames.Encoding, _encoding.WebName }
+                }
+            };
+
+            if (_operationId != null)
+            {
+                message.CorrelationId = _operationId;
+            }
```
fgheysels, line 141: Do we have this standard behavior somewhere documented ?

I'm not saying this is wrong; just wondering how we decided that the operationId should be used to set the CorrelationId, and why for instance we do not use the transactionId for that ?

I was just reading up on [this](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messages-payloads#message-routing-and-correlation), hence my question

stijnmoreels: In our docs, yeah, maybe not that explicitly. This functionality was previously in the `ObjectExtension` but it felt like that extension was becoming to big and complex to be readable, so this message build is a alternative. When I read it correctly in the Microsoft docs, the `CorrelationId` is closely linked to the message itself, which is the exact purpose of the operation ID, and not the transaction ID. It will be a good idea, indeed, to more explicitly document this message builder and how the correlation is set.

fgheysels: Fine for me.  I was just wondering if there should be a way for users to customize / override this behavior.

fgheysels (approved):

src/Arcus.Messaging.ServiceBus.Core/ServiceBusMessageBuilder.cs
stijnmoreels: Yeah, that's definitely a good point. Maybe we can provide an overload where they can provide an property name, so in that case the operation ID is (like the transaction ID and operation parent ID) assigned to the application properties of the message. Would that be ok?

stijnmoreels: I'll rename the method to `WithOperationId` and created a follow-up issue: #279 .

PR #284, opened 2022-05-20
chore: add package readme project tag

stijnmoreels:
Adds package readme project tag in project files.

Relates to https://github.com/arcus-azure/arcus/issues/209


fgheysels (approved):

fgheysels:
Looks like some check is stuck, but I don't see how I can re-trigger it.  Or do we have to re-run all checks ?

stijnmoreels:
> Looks like some check is stuck, but I don't see how I can re-trigger it. Or do we have to re-run all checks ?

It's the semantic check, we can override it, no problem. 😄

PR #295, opened 2022-06-06
fix: relax correlation properties guards in message builder

### LINKED ISSUE #293: ServiceBusMessageBuilder should be more tolerant for unspecified arguments
The `ServiceBusMessageBuilder` type has methods like: 
`WithOperationId`
`WithTransactionId`
etc...

These methods throw an exception if you pass in an empty or null operationId, transactionId, etc...
This means that the caller needs to check upfront if he can call that message, which makes the calling code much more bloated.
It would mean that you cannot write code like this:

image

because, whenever one of the id's is not available, it will throw.

Next to that, it will also result in exceptions being thrown and messages that cannot be sent, just for this.  
Logging is great, but if you run into a situation where you would normally add an operation-id, and that id is missing for some reason or another, we should not cause that a message cannot be sent.   Businesses rely on messaging, so we should not throw there.

When the `operationId`, `parentId`, ... is not there, we should just not add that information in the application-properties of the message.   I think it is also the power of such a builder pattern, that it takes care of those things.

stijnmoreels:
Relaxes the guards for correlation properties in the Azure Service Bus message builder.

Closes #293

### DIFF
--- a/src/Arcus.Messaging.Tests.Unit/Fixture/Order.cs
+++ b/src/Arcus.Messaging.Tests.Unit/Fixture/Order.cs
@@ -7,10 +7,69 @@ namespace Arcus.Messaging.Tests.Unit.Fixture
     /// <summary>
     /// Test message implementation to test specific message handling scenario's.
     /// </summary>
-    public class Order
+    public class Order : IEquatable<Order>
     {
         public string OrderId { get; set; }
 
         public string CustomerName { get; set; }
+
+        /// <summary>Indicates whether the current object is equal to another object of the same type.</summary>
+        /// <param name="other">An object to compare with this object.</param>
+        /// <returns>
+        /// <see langword="true" /> if the current object is equal to the <paramref name="other" /> parameter; otherwise, <see langword="false" />.</returns>
+        public bool Equals(Order other)
+        {
+            if (other is null)
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, other))
+            {
+                return true;
+            }
+
+            return OrderId == other.OrderId && CustomerName == other.CustomerName;
+        }
+
+        /// <summary>
+        /// Determines whether the specified object is equal to the current object.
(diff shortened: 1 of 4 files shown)


src/Arcus.Messaging.Tests.Unit/Fixture/Order.cs
```diff
@@ -7,10 +7,69 @@ namespace Arcus.Messaging.Tests.Unit.Fixture
     /// <summary>
     /// Test message implementation to test specific message handling scenario's.
     /// </summary>
-    public class Order
+    public class Order : IEquatable<Order>
```
fgheysels, line 10: If you specify `Order` to be a `record` instead of a `class`, you do not have to worry anymore about implementing IEquatable, Equals, etc... :)

fgheysels (approved):

src/Arcus.Messaging.Tests.Unit/Fixture/Order.cs
stijnmoreels: Ah, we could do that, indeed. (We could also use F# 😄 ).

stijnmoreels: That it is a test project doesn't really matter here, as it was about the general Arcus contribution that I was talking about 😄 . It is a way of learning, that's true.
Let me check if it reacts as expected with the other tests, as it is used quite a lot.

stijnmoreels: Ah, oops, we test both .NET Core and .NET 6 in the unit tests, so it would only work for .NET 6. We could however come back to this if we remove .NET Core from Arcus.

fgheysels: Is it ?  Isn't it a language feature that is not coupled with a .NET version ?

PR #304, opened 2022-06-17
feat: use message correlation model directly

### LINKED ISSUE #301: Assign message correlation information from model, not accessor
Currently, we're taking a detour when assigning the message correlation from the message router to the Serilog `LogContext`. First the correlation accessor is assigned, which gets retrieved back when the logs are written. This caused a problem as the request tracking seems to late in the chain and misses out on the correlation assignment.

Use the passed along correlation model in the message router to directly assign it in the Serilog `LogContext`. We can still use the `MessageCorrelationInfoEnricher` but it should take in the correlation model instead of the accessor.

stijnmoreels:
Use the message correlation model directly instead of doing a detour via the message correlation accessor.
This will make sure that we have the correct message correlation is used and will be more efficient and stable instead of going through the service provider.

Closes #301

### DIFF
--- a/src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
+++ b/src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
@@ -8,6 +8,7 @@ using Arcus.Observability.Telemetry.Core;
 using Arcus.Messaging.Abstractions.Telemetry;
 using Azure.Messaging.ServiceBus;
 using GuardNet;
+using Microsoft.Extensions.Azure;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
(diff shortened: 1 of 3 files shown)


src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
```diff
@@ -208,10 +209,13 @@ protected AzureServiceBusMessageRouter(IServiceProvider serviceProvider, AzureSe
             var isSuccessful = false;
             using (DurationMeasurement measurement = DurationMeasurement.Start())
             using (IServiceScope serviceScope = ServiceProvider.CreateScope())
-            using (UsingMessageCorrelationEnricher(serviceScope.ServiceProvider, correlationInfo))
```
fgheysels, line 211: What is this `UsingMessageCorrelationEnricher` method and should we remove it, as it seems that we're no longer using it ?

stijnmoreels: It was previously a returned `IDisposable` of Serilog's `LogContext`. It was `protected`, so made it deprecated.

fgheysels (approved):

PR #306, opened 2022-06-20
feat: make message correlation enrichment properties configurable

### LINKED ISSUE #303: Make message correlation Serilog property names configurable
Currently, we don't support custom correlation properties in the log events, while this is supported in the HTTP variant.

We should make sure that we support custom Serilog property values in the message correlation enricher, so people can control the correlation settings in both the Application Insights sink as well as with the enrichments (both HTTP and Messaging).

stijnmoreels:
Make the Serilog log event properties that are being enriched from the message correlation model configurable, as they were using these constants `ContextProperties.xxx` previously.

Closes #303

### DIFF
--- a/src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouterOptions.cs
+++ b/src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouterOptions.cs
@@ -21,6 +21,6 @@ namespace Arcus.Messaging.Abstractions.MessageHandling
         /// <summary>
         /// Gets the options to control the Serilog <see cref="MessageCorrelationInfoEnricher"/> when the incoming message is routed via the message router.
         /// </summary>
-        public MessageCorrelationEnricherOptions Telemetry { get; } = new MessageCorrelationEnricherOptions();
+        public MessageCorrelationEnricherOptions CorrelationEnricher { get; } = new MessageCorrelationEnricherOptions();
     }
 }
--- a/src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouterOptions.cs
+++ b/src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouterOptions.cs
@@ -1,4 +1,7 @@
-﻿namespace Arcus.Messaging.Abstractions.MessageHandling
+﻿using Arcus.Messaging.Abstractions.Telemetry;
+using Arcus.Observability.Telemetry.Serilog.Enrichers.Configuration;
+
+namespace Arcus.Messaging.Abstractions.MessageHandling
 {
     /// <summary>
     /// Represents the consumer-configurable options to change the behavior of the <see cref="MessageRouter"/>.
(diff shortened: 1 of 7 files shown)


src/Arcus.Messaging.Pumps.ServiceBus/Configuration/IAzureServiceBusQueueMessagePumpOptions.cs
```diff
@@ -47,11 +48,13 @@ public interface IAzureServiceBusQueueMessagePumpOptions
         /// <summary>
         /// Gets the options to control the correlation information upon the receiving of Azure Service Bus messages in the <see cref="AzureServiceBusMessagePump"/>.
         /// </summary>
+        [Obsolete("Will be moved to message routing options in the future")]
```
stijnmoreels, line 51: The problem here is that we have interfaces to expose subsections of the options to users (like: differentiate between queues and topics), but these interfaces should be updated with the message routing instead of (currently) the message routing properties.

So, now, we have `MessageRouterOptions.Correlation` but we expose it in the interfaces as `Correlation`. If we in the next major add a `Routing` property to the interfaces, it will be more maintainable for us.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouterOptions.cs
```diff
@@ -14,5 +17,10 @@ public class MessageRouterOptions
         /// Gets the options to control the correlation information upon the receiving of messages in the message router.
         /// </summary>
         public MessageCorrelationOptions Correlation { get; } = new MessageCorrelationOptions();
+
+        /// <summary>
+        /// Gets the options to control the Serilog <see cref="MessageCorrelationInfoEnricher"/> when the incoming message is routed via the message router.
+        /// </summary>
+        public MessageCorrelationEnricherOptions Telemetry { get; } = new MessageCorrelationEnricherOptions();
```
stijnmoreels, line 24: Not sure about the name though....

fgheysels: Telemetry is indeed a weird name, as the property defines some kind of options.
Why not `CorrelationEnricherOptions` instead of `Telemetry`?

fgheysels: To be consistent, we would then also need to rename `Correlation` to `CorrelationOptions` ?
Or rename the `Telemetry` property to `CorrelationEnricher` ?

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/IAzureServiceBusQueueMessagePumpOptions.cs
```diff
@@ -47,11 +48,13 @@ public interface IAzureServiceBusQueueMessagePumpOptions
...
         AzureServiceBusCorrelationOptions Correlation { get; }
         
         /// <summary>
         /// Gets the consumer-configurable options to change the deserialization behavior.
         /// </summary>
+        [Obsolete("Will be moved to message routing options in the future")]
```
fgheysels, line 57: Message routing options or the MessageRoutingOptions ?

stijnmoreels: Well, it will probably be called `Routing` or `MessageRouting` to be consistent, but since it's not available yet in the interface (only in the implementation), I wouldn't want to fixate on the name in the deprecated message.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouterOptions.cs
stijnmoreels: It will be good bc it shows the relation with the `Correlation` option. Good suggestion! 👍

fgheysels (approved):

PR #312, opened 2022-07-07
feat: add servicebus sender correlation extensions

### LINKED ISSUE #309: Add message correlation to Azure Service Bus client
Currently, users are responsible for their own creation of the Service Bus message, setting the correct correlation information, and placing that information in the tracking of the a Service Bus dependency.
Just like we helped with the HTTP tracking, we could do better and register our own wrapper for the Azure client which takes care of correlation under the hood.

Create an extension for the `AddServiceBusClient` which creates a new client with message correlation.

**Additional context**
https://docs.microsoft.com/en-us/dotnet/api/overview/azure/microsoft.extensions.azure-readme-pre

stijnmoreels:
Add extensions on the Azure Service Bus `ServiceBusSender` model to more easily send and track correlated messages that support service-to-service correlation.

Closes #309

### DIFF
--- a/docs/preview/02-Features/04-service-bus-extensions.md
+++ b/docs/preview/02-Features/04-service-bus-extensions.md
@@ -1,17 +1,11 @@
----
+﻿---
 title: "Azure Service Bus Extensions"
 layout: default
 ---
 
 # Azure Service Bus Extensions
 
-We provide several additional features related to message creation and message/context discoverability.
-
-- [Azure Service Bus Extensions](#azure-service-bus-extensions)
-  - [Installation](#installation)
-  - [Simplify Creating Service Bus Messages](#simplify-creating-service-bus-messages)
-  - [Simplify Message Information Discovery](#simplify-message-information-discovery)
-  - [Simplify Message Context Information Discovery](#simplify-message-context-information-discovery)
+We provide several additional features related to message creation/sending and message/context discoverability.
 
 ## Installation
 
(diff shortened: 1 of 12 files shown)


fgheysels (commented):
I thought we're going to implement this with some kind of factory cfr how we do it in the service-to-service correlation poc ?

stijnmoreels:
> I thought we're going to implement this with some kind of factory cfr how we do it in the service-to-service correlation poc ?

Yes, problem is that the Azure SDK doesn't allow enough 'hooks' for us to create an dedicated `ServiceBusSender` with all the things required. (mentioned here: https://github.com/arcus-azure/arcus.messaging/issues/309). So we are left with the same idea that we also support with the `HttpClient`, and that is extensions.

stijnmoreels:
> I thought we're going to implement this with some kind of factory cfr how we do it in the service-to-service correlation poc ?

The poc uses the Azure client services to inject an `ServiceBusSender`, but still does the correlation by itself. Shame that we can't hook enough stuff here to do our correlation tracking.

fgheysels:
> > I thought we're going to implement this with some kind of factory cfr how we do it in the service-to-service correlation poc ?
> 
> The poc uses the Azure client services to inject an `ServiceBusSender`, but still does the correlation by itself. Shame that we can't hook enough stuff here to do our correlation tracking.
> 
> * The `.AddServiceBusClient` doesn't return an extensible type, only a marker interface,
> * The `ServiceBusClient.CreateSender` can be overriden, but the `ServiceBusSender` itself is not able to be overriden

I was more referring to the `ServiceMessageBuilder` construct, but as I understand it, we already support this (ServiceMessageBuilder) and this is just an additional way for making sure messages are send with the correct correlation options ?

stijnmoreels:
> I was more referring to the `ServiceMessageBuilder` construct, but as I understand it, we already support this (ServiceMessageBuilder) and this is just an additional way for making sure messages are send with the correct correlation options ?

Not quite, as these extensions also make sure that the dependency is logged correctly both with the assignment of the dependency ID as well as the duration the dependency call took.
But yes, since we also support the builder, consumers are free to use any one of those.

fgheysels:
> > I was more referring to the `ServiceMessageBuilder` construct, but as I understand it, we already support this (ServiceMessageBuilder) and this is just an additional way for making sure messages are send with the correct correlation options ?
> 
> Not quite, as these extensions also make sure that the dependency is logged correctly both with the assignment of the dependency ID as well as the duration the dependency call took. But yes, since we also support the builder, consumers are free to use any one of those.

I see, with the extension method, the call to the servicebus dependency is logged as well, whereas with the builder, you just create the message with all required properties set.  However, you're still responsible for making sure the dependency is logged.

fgheysels (approved):

stijnmoreels:
> > > I was more referring to the `ServiceMessageBuilder` construct, but as I understand it, we already support this (ServiceMessageBuilder) and this is just an additional way for making sure messages are send with the correct correlation options ?
> > 
> > 
> > Not quite, as these extensions also make sure that the dependency is logged correctly both with the assignment of the dependency ID as well as the duration the dependency call took. But yes, since we also support the builder, consumers are free to use any one of those.
> 
> I see, with the extension method, the call to the servicebus dependency is logged as well, whereas with the builder, you just create the message with all required properties set. However, you're still responsible for making sure the dependency is logged.

Yes, that's correct! 👍

Issue #24, opened 2019-12-04
Provide basic message pump for Azure Event Hubs


tomkerkhove:
We should provide a similar message pump which supports Azure Event Hubs and shares the core of the logic with Azure Service Bus which was added is being added via #14.

We need to verify if:
- We can use an open-standard/protocol (AMQP?) to process the messages with to hopefully support pulling from Kafka as well, if possible
- This will require checkpointing of processed messages in the stream, how will we approach this and what will we support

We need to build this on top of [Event Processor Host](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-event-processor-host) so that it handles the partition determination with Azure Storage

PR #323, opened 2022-07-26
feat: add eventhubs messagehandler overloads

stijnmoreels:
Adds message handler overloads for the Azure EventHubs message pump/router.

Relates to #24


fgheysels (approved):
I think this looks good!

PR #334, opened 2022-08-30
fix: use general correlation model in sender extensions

stijnmoreels:
Changes the messaging-specific correlation model `MessageCorrelationInfo` with the general `CorrelationInfo` so that application from other environments (ex. HTTP) can also benefit from these extensions without having to convert to the messaging-specific model.


stijnmoreels:
We can still do this bc it is not yet released.

fgheysels (approved):
That was easy to review !

PR #338, opened 2022-09-14
feat: generalize restartable message pumps

### LINKED ISSUE #333: Generalize restart functionality in message pumps with interface so it can be generally restarted
The background jobs functionality allows for message pumps to be restarted based on rotated credentials. With the addition of the Azure EventHubs message pump, such restart functionality could be useful there, too.

Instead of hard-coding it on the Azure Service Bus message pump, we should expose something like a 'restart interface' that we can re-use across background jobs that want to restart a message pump. The background job registration then only needs the job ID and to identify which message pump it needs to restart, instead of now be hard-coded on the type of message pump it can restart.

stijnmoreels:
Generalize restart functionality so that background jobs listening on events can more easily restart message pumps.

Closes #333

### DIFF
--- a/src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandlerCollection.cs
+++ b/src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandlerCollection.cs
@@ -20,6 +20,12 @@ namespace Arcus.Messaging.Abstractions.MessageHandling
             Guard.NotNull(services, nameof(services), "Requires a collection of services to register the message handling logic into");
             Services = services;
         }
+
+        /// <summary>
+        /// Gets or sets the unique message pump ID to identify the message pump throughout the application.
+        /// This ID can be used to get a reference of the previously registered message pump for which a message handler is registered.
+        /// </summary>
+        public string JobId { get; set; }
         
         /// <summary>
         /// Gets the current available collection of services to register the message handling logic into.
--- a/src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandlerCollection.cs
+++ b/src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandlerCollection.cs
@@ -22,8 +22,8 @@ namespace Arcus.Messaging.Abstractions.MessageHandling
         }
 
         /// <summary>
-        /// Gets or sets the unique message pump ID to identify the message pump throughout the application.
-        /// This ID can be used to get a reference of the previously registered message pump for which a message handler is registered.
+        /// Gets or sets the unique ID to identify the job that runs a message pump during the lifetime of the application.
+        /// This ID can be used to get a reference of the previously registered message pump while registering message handlers and other functionality related to the message pump.
         /// </summary>
         public string JobId { get; set; }
         
--- a/src/Arcus.Messaging.Pumps.Abstractions/IRestartableMessagePump.cs
+++ b/src/Arcus.Messaging.Pumps.Abstractions/IRestartableMessagePump.cs
@@ -1,0 +1,23 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+namespace Arcus.Messaging.Pumps.Abstractions
+{
+    /// <summary>
+    /// Represents an <see cref="MessagePump"/> that can be restarted programmatically.
+    /// </summary>
+    public interface IRestartableMessagePump : IHostedService
+    {
+        /// <summary>
+        /// Gets the unique message pump ID to identify the pump that needs to be restarted.
+        /// </summary>
+        public string JobId { get; }
+
+        /// <summary>
+        /// Programmatically restart the message pump.
+        /// </summary>
+        /// <param name="cancellationToken">The token to cancel the restart process.</param>
+        Task RestartAsync(CancellationToken cancellationToken);
+    }
+}
--- a/src/Arcus.Messaging.Pumps.Abstractions/IRestartableMessagePump.cs
+++ b/src/Arcus.Messaging.Pumps.Abstractions/IRestartableMessagePump.cs
@@ -12,7 +12,7 @@ namespace Arcus.Messaging.Pumps.Abstractions
         /// <summary>
         /// Gets the unique message pump ID to identify the pump that needs to be restarted.
         /// </summary>
-        public string JobId { get; }
+        string JobId { get; }
 
         /// <summary>
         /// Programmatically restart the message pump.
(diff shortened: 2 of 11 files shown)


src/Arcus.Messaging.Pumps.Abstractions/IRestartableMessagePump.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Represents an <see cref="MessagePump"/> that can be restarted programmatically.
+    /// </summary>
+    public interface IRestartableMessagePump : IHostedService
+    {
+        /// <summary>
+        /// Gets the unique message pump ID to identify the pump that needs to be restarted.
+        /// </summary>
+        public string JobId { get; }
```
stijnmoreels, line 15:
```suggestion
        string JobId { get; }
```

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandlerCollection.cs
```diff
@@ -20,6 +20,12 @@ public MessageHandlerCollection(IServiceCollection services)
             Guard.NotNull(services, nameof(services), "Requires a collection of services to register the message handling logic into");
             Services = services;
         }
+
+        /// <summary>
+        /// Gets or sets the unique message pump ID to identify the message pump throughout the application.
+        /// This ID can be used to get a reference of the previously registered message pump for which a message handler is registered.
+        /// </summary>
+        public string JobId { get; set; }
```
fgheysels, line 28: I don't see why this property is called `JobId` but in the comments it is explained as to be the Message Pump ID.
(And then, I also do not see how this property (JobId / Pump ID) is set on a collection of message handlers ?

src/Arcus.Messaging.Pumps.Abstractions/IRestartableMessagePump.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Represents an <see cref="MessagePump"/> that can be restarted programmatically.
+    /// </summary>
+    public interface IRestartableMessagePump : IHostedService
+    {
+        /// <summary>
+        /// Gets the unique message pump ID to identify the pump that needs to be restarted.
+        /// </summary>
+        string JobId { get; }
```
fgheysels, line 15: Again, the property is called `JobId`, but it is a member of a MessagePump class, and in the comments it is also refered to as 'message pump id'.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandlerCollection.cs
stijnmoreels: If we set this ID when registering the message pump, extensions can make use of this pump. The message handlers are registered on the application services directly and may not use this, but a background job that wants to extend the message pump needs such an ID to know for which message pump it should run the background job. (Think about the background job that restarts message pumps on rotated credentials). If we pass this job ID, we don't need the consumer to pass it themselves.
This is also somewhat explained in the related issue here: https://github.com/arcus-azure/arcus.messaging/issues/333

fgheysels: Should we then call it MessagePumpId, as you are identifying a message pump ?

src/Arcus.Messaging.Pumps.Abstractions/IRestartableMessagePump.cs
stijnmoreels: We already had a `JobId` as a property on the pump that was identifying it as such. The reason we choose 'job' is probably to identify any type of hosted service. Also, some kind of background job. We also have already a `Id` property on the message pump (not sure why, honestly), so if we would add another one, it would maybe make this more complicated, no?

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandlerCollection.cs
stijnmoreels: We already had a `JobId` as a property on the pump that was identifying it as such. The reason we choose 'job' is probably to identify any type of hosted service. Also, some kind of background job. We also have already a `Id` property on the message pump (not sure why, honestly), so if we would add another one, it would maybe make this more complicated, no?

fgheysels: It's still not clear to me :)  Maybe I should use it to get a hold of it.
But if you say we already had a JobId as a property on the pump, then it means that  this  existing property is just promoted to be an interface member ?

stijnmoreels: > But if you say we already had a JobId as a property on the pump, then it means that this existing property is just promoted to be an interface member ?

Yes, that's correct. 👍

fgheysels: Ok, so it's not that easy to change the name of it, as it refers to an existing property.
I'm not entirely convinced regarding the name, and the comment that describes the property adds a bit to the confusion tbh. Maybe the comment can be improved ?

stijnmoreels: Sure, we can do that. Any suggestions? As the first line of the comment is exactly the purpose of the property. Maybe it is the second line that adds confusion?

fgheysels: The 2nd line is indeed confusing.
The 1st line is also somewhat confusing as the property is named 'JobId', while in the docs it is explained to be the Message Pump ID

stijnmoreels: Changed it to:
```csharp
/// Gets or sets the unique ID to identify the job that runs a message pump during the lifetime of the application.
/// This ID can be used to get a reference of the previously registered message pump while registering message handlers and other functionality related to the message pump.
```

src/Arcus.Messaging.Pumps.EventHubs/AzureEventHubsMessagePump.cs
```diff
@@ -55,13 +55,20 @@ internal class AzureEventHubsMessagePump : MessagePump
...
         }
 
         private string EventHubName => _eventProcessor?.EventHubName;
         private string ConsumerGroup => _eventProcessor?.ConsumerGroup;
         private string Namespace => _eventProcessor?.FullyQualifiedNamespace;
 
+        /// <summary>
+        /// Gets the unique message pump ID to identify the pump that needs to be restarted.
+        /// </summary>
+        public string JobId { get; }
```
fgheysels, line 70: Should we use `<inheritdoc/>` in the comments section ?
(For all implemented interface members ? )

fgheysels (approved):
I just still have one little remark :)

src/Arcus.Messaging.Pumps.EventHubs/AzureEventHubsMessagePump.cs
stijnmoreels: https://github.com/arcus-azure/arcus.messaging/pull/338/commits/49b8bea3e8ab54c2f1e4a57a76fa0a85f4803bb7

PR #344, opened 2022-10-19
feat: add w3c correlation support

### LINKED ISSUE #342: Introduce W3C correlation tracking with automatic dependency tracking
Currently, any built-in Microsoft dependencies in the message handlers implementations are not automatically tracked without Arcus extensions.

Make use of the `Diagnostic-Id` application property which is the `traceparent` of the W3C correlation relationship instead of custom application properties. That way, both Arcus and built-in Microsoft dependencies are tracked with the same parent.

stijnmoreels:
Add W3C correlation support to include automatic (Microsoft) dependency tracking.

Closes #342


src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Functions/OrderProcessingFunction.cs
```diff
@@ -31,10 +36,14 @@ public OrderProcessingFunction(IAzureServiceBusMessageRouter messageRouter)
...
-            MessageCorrelationInfo correlationInfo = message.GetCorrelationInfo();
-            await _messageRouter.RouteMessageAsync(message, context, correlationInfo, cancellationToken);
+
+            AzureServiceBusMessageContext context = message.GetMessageContext(_jobId);
+            (string transactionId, string operationParentId) = message.ApplicationProperties.GetTraceParent();
+
+            using (var result = MessageCorrelationResult.Create(_telemetryClient, transactionId, operationParentId))
+            {
+                await _messageRouter.RouteMessageAsync(message, context, result.CorrelationInfo, cancellationToken);
+            }
```
stijnmoreels, line 46: Will be migrated towards an isolated environment in https://github.com/arcus-azure/arcus.messaging/issues/346

stijnmoreels:
I'll let you already review this part 😉 .

src/Arcus.Messaging.Abstractions/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -64,6 +64,8 @@ public static MessageHandlerCollection AddMessageRouting(this IServiceCollection
             Guard.NotNull(services, nameof(services), "Requires a set of services to add the message routing");
             Guard.NotNull(implementationFactory, nameof(implementationFactory), "Requires a function to create the message router");
 
+            services.AddApplicationInsightsTelemetryWorkerService();
```
fgheysels, line 67: Just wondering why we have the `WorkerService` part in the methodname ?

src/Arcus.Messaging.Abstractions/MessageCorrelationResult.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+
+namespace Arcus.Messaging.Abstractions
+{
+    /// <summary>
+    /// Represents the correlation result of a received Azure Service Bus message.
+    /// This result will act as the scope of the request telemetry.
+    /// </summary>
```
fgheysels, line 13: If I'm not mistaken, this class is only useful when using hierarchical logging ?
Should we add this as a `<remark>` ?

src/Arcus.Messaging.Pumps.EventHubs/AzureEventHubsMessagePump.cs
```diff
@@ -142,13 +145,28 @@ private async Task ProcessMessageAsync(ProcessEventArgs args)
 
             // TODO: create extension on `EventData` to retrieve context.
```
fgheysels, line 146: Is this todo on the backlog ?
If not, we can add it to the backlog and remove this 'todo' ?

src/Arcus.Messaging.Tests.Integration/Fixture/Logging/InMemoryTelemetryChannel.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.ApplicationInsights.Channel;
+
+namespace Arcus.Messaging.Tests.Integration.Fixture.Logging
+{
+    public class InMemoryTelemetryChannel : ITelemetryChannel
```
fgheysels, line 11: Is this an interface that we control / own ? (ITelemetryChannel)

src/Arcus.Messaging.Tests.Integration/MessagePump/Fixture/AssertX.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.Extensions.Logging;
+using Polly;
+using Polly.Retry;
+using Xunit;
+
+namespace Arcus.Messaging.Tests.Integration.MessagePump.Fixture
+{
+    public static class AssertX
```
fgheysels, line 12: Why AssertX ?  Just wondering :)

src/Arcus.Messaging.Tests.Workers.EventHubs.Core/Arcus.Messaging.Tests.Workers.EventHubs.Core.csproj
```diff
@@ -11,6 +11,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.0" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="6.0.0" />
     <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.9.5" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.8.4" />
```
fgheysels, line 14: Isn't Microsoft.SqlClient the new package ?

src/Arcus.Messaging.Abstractions/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: This is an extension from Microsoft. This will make sure that the `TelemetryClient` is registered and ready to be used when built-in Microsoft dependencies are being tracked in custom message handlers.

src/Arcus.Messaging.Abstractions/MessageCorrelationResult.cs
stijnmoreels: That's incorrect. Both for W3C as Hierarchical logging this class is used, so as to avoid confusing and have a streamliend approach.

src/Arcus.Messaging.Pumps.EventHubs/AzureEventHubsMessagePump.cs
stijnmoreels: Yes, added it recently: https://github.com/arcus-azure/arcus.messaging/issues/345
Will remove TODO. 👍

src/Arcus.Messaging.Tests.Integration/MessagePump/Fixture/AssertX.cs
stijnmoreels: Because to remove conflict from xUnit `Assert` class. (Oh, how I which we were in F# 😉, with their automatic module merging). We did this on other Arcus repo's too.

src/Arcus.Messaging.Tests.Workers.EventHubs.Core/Arcus.Messaging.Tests.Workers.EventHubs.Core.csproj
stijnmoreels: Ah, not sure. Was using this package in our Web API W3C stuff, too.

src/Arcus.Messaging.Pumps.EventHubs/AzureEventHubsMessagePump.cs
stijnmoreels: Removed in https://github.com/arcus-azure/arcus.messaging/pull/344/commits/4b53213f075c3bda880746c1bd257439c50bbe8d

src/Arcus.Messaging.Tests.Workers.EventHubs.Core/Arcus.Messaging.Tests.Workers.EventHubs.Core.csproj
fgheysels: I think we should use the Microsoft.Data.SqlClient package
https://learn.microsoft.com/en-us/sql/connect/ado-net/introduction-microsoft-data-sqlclient-namespace?view=sql-server-ver16

src/Arcus.Messaging.Abstractions/MessageCorrelationResult.cs
```diff
@@ -0,0 +1,97 @@
...
+    public class MessageCorrelationResult : IDisposable
+    {
+        private readonly TelemetryClient _telemetryClient;
```
fgheysels, line 16: TelemetryClient is an Application Insights specific class, so I think it is a bit weird to see it here in a namespace ethat is called 'Abstractions', and should not be coupled to App Insights ?

src/Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue/Functions/OrderProcessingFunction.cs
```diff
@@ -13,14 +14,18 @@ namespace Arcus.Messaging.Tests.Runtimes.AzureFunction.ServiceBus.Queue.Function
...
         /// <summary>
         /// Initializes a new instance of the <see cref="OrderProcessingFunction" /> class.
         /// </summary>
-        public OrderProcessingFunction(IAzureServiceBusMessageRouter messageRouter)
+        public OrderProcessingFunction(
+            IAzureServiceBusMessageRouter messageRouter,
+            TelemetryClient telemetryClient)
         {
             _messageRouter = messageRouter;
+            _telemetryClient = telemetryClient;
```
fgheysels, line 28: I think we should try to avoid to have telemetryClient instances here ?

stijnmoreels: See comment here: https://github.com/arcus-azure/arcus.messaging/pull/344#discussion_r1000168182
Will be extracted from the function but was too much for this PR to also include this change. Not that this adaption is only for in-process Azure Functions. We will probably have to come up with something similar as with the HTTP trigger where we have an `HttpBasedFunction` base class that hides these things, or an additional type that hides the client.

src/Arcus.Messaging.Tests.Workers.EventHubs.Core/Arcus.Messaging.Tests.Workers.EventHubs.Core.csproj
stijnmoreels: Yes, will change this.

src/Arcus.Messaging.Abstractions/MessageCorrelationResult.cs
stijnmoreels: The Observability library is not coupled to Application Insights, but we already have made that coupling in the W3C HTTP correlation support as Web API (and Messaging) are seen as 'implementations' of Observability.
If we want to include automatic dependency tracking for built-in Microsoft dependencies, without the additional configurations and publicly exposed extensions, and without action from the consumer, we have to include a reference to Application Insights without our 'implementations' of Observability. (like we did in Web API).
Also note that 'abstractions' here means 'messaging abstractions' as in these types are not coupled on a messaging system (message pump or Azure Functions).

src/Arcus.Messaging.Tests.Workers.EventHubs.Core/Arcus.Messaging.Tests.Workers.EventHubs.Core.csproj
stijnmoreels: Updated in https://github.com/arcus-azure/arcus.messaging/pull/344/commits/7ae36db2b1f43d6a0d256633c1018544bb3c4a0e

src/Arcus.Messaging.Abstractions/MessageCorrelationResult.cs
fgheysels: But using this TelemetryClient class means that we have a dependency on ApplicationInsights in this assembly.   I thought the only place were we had this dependency, was in the ApplicationInsights sink project.

stijnmoreels: No, we can't do that, as there are different Application Insights packages needed to do request/dependency tracking. The Web API needs the `AspNetCore` sub-package and the message pump and all other non-HTTP stuff needs `WorkerService`. The dependency tracking module is available via the dependency collector project, but all necessary boilerplate stuff is around. This module needs the telemetry configuration which means that a registration needs to be configured, this registration can either be done via the AspNetCore or WorkerService package. If we would try to register all the necessary stuff ourselves, we would not only make it very unstable for change, we could also break functionality if people want to extend Application Insights configuration.

stijnmoreels: Yes, I know. It's a shame. Don't see a way to do better than this at the moment.

fgheysels: OK. Let's put our quest for purity aside and be pragmatic. I think the value of auto-tracking is high enough to make this concession.

fgheysels (approved):

PR #362, opened 2022-11-21
feat: add servicebus client secretstore extension overload

### LINKED ISSUE #358: Add extensions overloads for Azure clients to use Arcus secret store
Currently, one has to pass in the connection string directly or via an `IConfiguration` instance if one wants to register an Azure ServiceBus client to the application.

Add an extension overload that lets you pass in a secret name which will be used when contacting the secret store.

stijnmoreels:
Add extension overload to register a `ServiceBusClient` via a secret name corresponding with the Azure Service Bus connection string that is registered in the Arcus secret store.

Closes #358

### DIFF
--- a/docs/preview/02-Features/04-service-bus-extensions.md
+++ b/docs/preview/02-Features/04-service-bus-extensions.md
@@ -15,7 +15,7 @@ This features requires to install our NuGet package:
 PM > Install-Package Arcus.Messaging.ServiceBus.Core
 ```
 
-## Using Arcus secret store when Service Bus client
+## Using Arcus secret store when registering the Service Bus client
 
 When registering a `ServiceBusClient` via [Azure's client registration process](https://learn.microsoft.com/en-us/dotnet/api/overview/azure/messaging.servicebus-readme), the library provides an extension to pass-in a secret name instead of directly passing the Azure Service Bus connection string.
 This secret name will correspond with a registered secret in the [Arcus secret store](https://security.arcus-azure.net/features/secret-store) that holds the Azure Service Bus connection string.
--- a/docs/preview/02-Features/04-service-bus-extensions.md
+++ b/docs/preview/02-Features/04-service-bus-extensions.md
@@ -15,6 +15,32 @@ This features requires to install our NuGet package:
 PM > Install-Package Arcus.Messaging.ServiceBus.Core
 ```
 
+## Using Arcus secret store when Service Bus client
+
+When registering a `ServiceBusClient` via [Azure's client registration process](https://learn.microsoft.com/en-us/dotnet/api/overview/azure/messaging.servicebus-readme), the library provides an extension to pass-in a secret name instead of directly passing the Azure Service Bus connection string.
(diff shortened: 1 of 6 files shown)


src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                var secretProvider = serviceProvider.GetService<ISecretProvider>();
+                if (secretProvider is null)
+                {
+                    throw new InvalidOperationException(
+                        "Requires an Arcus secret store registration to retrieve the connection string to authenticate with Azure Service Bus while creating an Service Bus client instance," 
+                        + "please use the 'services.AddSecretStore(...)' or 'host.ConfigureSecretStore(...)' (https://security.arcus-azure.net/features/secret-store)");
+                }
+
+                string connectionString = secretProvider.GetRawSecretAsync(connectionStringSecretName).GetAwaiter().GetResult();
```
stijnmoreels, line 70: Still a bit skeptical about this part. We could create 'proxy' classes around the `ServiceBusClient`, `ServiceBusSender` and `ServiceBusReceiver` so that upon any asynchronous call, the connection string secret would be retrieved, but was wondering if that would be worth the effort AND if that would be stable enough to use it if one of these classes changes.

docs/preview/02-Features/04-service-bus-extensions.md
```diff
@@ -15,6 +15,32 @@ This features requires to install our NuGet package:
...
+using Microsoft.Extensions.DependencyInjection;
+
+public class Program
+{
+    public void ConfigureServices(IServiceCollection services)
+    {
+        // Adding Arcus secret store, more info: https://security.arcus-azure.net/features/secret-store
+        services.AddSecretStore(stores => stores.AddAzureKeyVaultWithManagedIdentity("https://my.vault.azure.net");
+
+        // Adding Service Bus client with secret in Arcus secret store.
```
fgheysels, line 36: Simple but nice & useful feature!

src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
fgheysels: I understand, but I do not see an easy way out of this, except maybe providing non-async methods for retrieving the secrets.

fgheysels (approved):

docs/preview/02-Features/04-service-bus-extensions.md
stijnmoreels: Thought so too! 👍

src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
stijnmoreels: The idea to do this sync secret retrieval was already in the past mentioned, here: https://github.com/arcus-azure/arcus.security/issues/259, but we didn't really go through with it (yet). I think bc of the effort it would take for the minor benefits.

PR #367, opened 2022-12-07
chore: update cancelled logging

### LINKED ISSUE #321: Catch `OperationCancelledException` in message pumps
Let's catch `OperationCancelledException` to control cancellation ourselves, and log it ourselves in the message pumps. We could check also if the `stoppingToken.IsCancelRequested` results to `true` before catching so we're sure that only cancellations are logged this way.

stijnmoreels:
Use better logging when message pumps are cancelled.

Closes #321


fgheysels (approved):
Small improvements can be valuable as well ;)

PR #373, opened 2022-12-16
chore: use sync secret retrieval in az clients registrations

### LINKED ISSUE #374: Provide an Async method for registering Azure clients
We currently have methods like `AddEventHubProducerClient` that take a secret-name as an argument.   These methods will use the registered `ISecretProvider` to retrieve the secret that is necessary to instantiate the client that we want  to add to the DI container.

In the documentation of the Arcus template I read this warning:

> The Azure EventHubs connection string will be retrieved synchronously, so make sure that you register your connection string secret in a secret provider that supports synchronous secret retrieval

This is something the end-user has no control over.  Suppose someone uses a secret provider that doesn't support sync secret retrieval, then the user is blocked ?
Should we consider adding a `AddEventHubProducerClientAsync` method which retrieves the secret in an async way ?  (Or some other solution for this  ?)

stijnmoreels:
Use synchronous secret retrieval functionality to retrieve the connection strings in the Azure client registrations.

Closes #374

### DIFF
--- a/src/Arcus.Messaging.EventHubs.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
+++ b/src/Arcus.Messaging.EventHubs.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
@@ -4,6 +4,9 @@ using Azure.Core.Extensions;
 using Azure.Messaging.EventHubs.Producer;
 using GuardNet;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Serilog.Core;
 
 // ReSharper disable once CheckNamespace
 namespace  Microsoft.Extensions.Azure
@@ -51,7 +54,7 @@ namespace  Microsoft.Extensions.Azure
         /// <remarks>
         ///   <para>
         ///     Make sure that the Arcus secret store is registered in the application before using this extension (<a href="https://security.arcus-azure.net/features/secret-store">more info</a>)
-        ///     as the Azure EventHubs connection string will be synchronously retrieved via the <paramref name="connectionStringSecretName"/>.
+        ///     as the Azure EventHubs connection string will be retrieved via the <paramref name="connectionStringSecretName"/>.
         ///   </para>
         ///   <para>
         ///     If the connection string is copied from the Event Hubs namespace, it will likely not contain the name of the desired Event Hub, which is needed.
@@ -90,7 +93,7 @@ namespace  Microsoft.Extensions.Azure
         /// <remarks>
         ///   <para>
         ///     Make sure that the Arcus secret store is registered in the application before using this extension (<a href="https://security.arcus-azure.net/features/secret-store">more info</a>)
-        ///     as the Azure EventHubs connection string will be synchronously retrieved via the <paramref name="connectionStringSecretName"/>.
+        ///     as the Azure EventHubs connection string will be retrieved via the <paramref name="connectionStringSecretName"/>.
         ///   </para>
         ///   <para>
         ///     If the connection string is copied from the Event Hub itself, it will contain the name of the desired Event Hub,
@@ -124,7 +127,7 @@ namespace  Microsoft.Extensions.Azure
         /// <remarks>
         ///   <para>
         ///     Make sure that the Arcus secret store is registered in the application before using this extension (<a href="https://security.arcus-azure.net/features/secret-store">more info</a>)
-        ///     as the Azure EventHubs connection string will be synchronously retrieved via the <paramref name="connectionStringSecretName"/>.
+        ///     as the Azure EventHubs connection string will be retrieved via the <paramref name="connectionStringSecretName"/>.
         ///   </para>
         ///   <para>
         ///     If the connection string is copied from the Event Hub itself, it will contain the name of the desired Event Hub,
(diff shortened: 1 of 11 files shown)


fgheysels (approved):

src/Arcus.Messaging.EventHubs.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
```diff
@@ -20,7 +23,7 @@ public static class AzureClientFactoryBuilderExtensions
         /// <remarks>
         ///   <para>
         ///     Make sure that the Arcus secret store is registered in the application before using this extension (<a href="https://security.arcus-azure.net/features/secret-store">more info</a>)
-        ///     as the Azure EventHubs connection string will be retrieved via the <paramref name="connectionStringSecretName"/>.
+        ///     as the Azure EventHubs connection string will be synchronously retrieved via the <paramref name="connectionStringSecretName"/>.
```
fgheysels, line 26: Are we going to change the implementation of the secret retrieval so that we fall back to async retrieval if no secret could be retrieved synchronously ?  If so, then shouldn't we remove this warning ?

stijnmoreels: This warning is mostly about the consumer that needs to register the Arcus secret store for this to work. There were some 'synchronously' words there still, so these are now removed.
Fallback on async is implemented here, but not sure if we should do that in Arcus Security, as we could have unexpected results of trying to run asynchronous code synchronously, right?

src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
```diff
@@ -61,19 +65,38 @@ public static class AzureClientFactoryBuilderExtensions
...
+                string connectionString = secretProvider.GetRawSecret(connectionStringSecretName);
+                return connectionString;
+            }
+            catch (Exception exception)
+            {
+                ILogger logger = 
+                    serviceProvider.GetService<ILogger<ServiceBusClient>>() 
+                    ?? NullLogger<ServiceBusClient>.Instance;
+
+                logger.LogWarning(exception, "Cannot synchronously retrieve Azure Service Bus connection string secret for '{SecretName}', fallback on asynchronously", connectionStringSecretName);
```
fgheysels, line 96: Why is this a warning ?  I would categorize this more as a trace.

src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
```diff
@@ -51,6 +54,7 @@ public static class AzureClientFactoryBuilderExtensions
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c>.</exception>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="connectionStringSecretName"/> is blank.</exception>
         /// <exception cref="InvalidOperationException">Thrown when the Arcus secret store is not registered.</exception>
+        /// <exception cref="NotSupportedException">Thrown when the registered Arcus secret store does not have a synchronous secret provider registered.</exception>
```
fgheysels, line 57: I don't think we throw an exception when no sync secret provider is registered ?  We're falling back to retrieve the secret async.

src/Arcus.Messaging.EventHubs.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
```diff
@@ -169,8 +176,21 @@ private static string GetEventHubsConnectionString(string connectionStringSecret
...
+                string connectionString = secretProvider.GetRawSecret(connectionStringSecretName);
+                return connectionString;
+            }
+            catch (Exception exception)
+            {
+                ILogger logger = 
+                    serviceProvider.GetService<ILogger<EventHubProducerClient>>() 
+                    ?? NullLogger<EventHubProducerClient>.Instance;
+
+                logger.LogWarning(exception, "Cannot synchronously retrieve Azure EventHubs connection string secret for '{SecretName}', fallback on asynchronously", connectionStringSecretName);
```
fgheysels, line 190: I would not log this as a warning, rather as a trace or debug.
If you do not have a secret provider registered that supports sync retrieval, then this is something that you're not going to 'fix'.  It's also not a real issue imho.

stijnmoreels: Moved to `LogTrace` in https://github.com/arcus-azure/arcus.messaging/pull/373/commits/efbcb36c397809e914f3951916188d0c51fc8344

src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
stijnmoreels: Aha, yes, removed this one and added `SecretNotFoundException`. 👍

src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
stijnmoreels: Yes, updated to `LogTrace` in https://github.com/arcus-azure/arcus.messaging/pull/373/commits/efbcb36c397809e914f3951916188d0c51fc8344

src/Arcus.Messaging.EventHubs.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
fgheysels: Yes that is right.

PR #386, opened 2023-01-20
chore: add code coverage (🏗️🚧👷‍♂️)

stijnmoreels:
Add code coverage reporting.

Relates to https://github.com/arcus-azure/arcus/issues/264

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -25,7 +25,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Coverlet coverage test
-      run: dotnet test -c Debug -p:coverletOutput=coverage.xml -p:CollectCoverage=true -p:CoverletOutputFormat=opencover -p:Threshold=80 -p:ExcludeByAttribute=\"Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute\" -p:Exclude=\"[Arcus.Messaging.Tests.*]*,[Arcus.Messaging.Pumps.*]*\" src/Arcus.Messaging.Tests.Unit/Arcus.Messaging.Tests.Unit.csproj
+      run: dotnet test -c Debug -p:coverletOutput=coverage.xml -p:CollectCoverage=true -p:CoverletOutputFormat=opencover -p:Threshold=80 -p:ExcludeByAttribute=\"Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute\" -p:Exclude=\"[Arcus.Messaging.Tests.*]*\" src/Arcus.Messaging.Tests.Unit/Arcus.Messaging.Tests.Unit.csproj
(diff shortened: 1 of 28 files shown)


stijnmoreels:
Excluded the pumps projects as these are tested in the integration tests and are mostly containing code to contact Azure resources a.k.a. integration stuff.

fgheysels:
> Excluded the pumps projects as these are tested in the integration tests and are mostly containing code to contact Azure resources a.k.a. integration stuff.

I understand the reasoning, but it feels a bit wrong to me.  It feels like it's a way to pimp the code-coverage results and although I know that they're being tested via integration tests, I don't think we should 'fix' this in this way.

Ideally we run code-coverage during integration tests as well,and merge those results together with the unit-test results into a single report.
However, I think the problem is here that integration tests are not always being run and unit-tests are ran by every pipeline-run.

We can maybe fix this by making sure that we only publish the code-coverage reports from the nightly build pipelines where the integration-tests are run as well.

stijnmoreels:
> However, I think the problem is here that integration tests are not always being run and unit-tests are ran by every pipeline-run.

> We can maybe fix this by making sure that we only publish the code-coverage reports from the nightly build pipelines where the integration-tests are run as well.

That's not entirely true. We have two build pipelines: CI and Release, and in both pipelines we run both the unit and integration tests.

The problem with merging the reports, is that the threshold functionality does not work as it still inspects the separate reports. The same problem occurred in the [Web API project](https://github.com/arcus-azure/arcus.webapi/pull/432). I think it will be the choice to either merge the reports and let go of the threshold, or only running the unit tests. 
We could also come up with a way to inspect the merged report ourselves, and look for the total code coverage percentage, but that should be investigated.

fgheysels:
> > However, I think the problem is here that integration tests are not always being run and unit-tests are ran by every pipeline-run.
> 
> > We can maybe fix this by making sure that we only publish the code-coverage reports from the nightly build pipelines where the integration-tests are run as well.
> 
> That's not entirely true. We have two build pipelines: CI and Release, and in both pipelines we run both the unit and integration tests.
> 
> The problem with merging the reports, is that the threshold functionality does not work as it still inspects the separate reports. The same problem occurred in the [Web API project](https://github.com/arcus-azure/arcus.webapi/pull/432). I think it will be the choice to either merge the reports and let go of the threshold, or only running the unit tests. We could also come up with a way to inspect the merged report ourselves, and look for the total code coverage percentage, but that should be investigated. Also, when we decide to run these integration tests, we should also [decide to migrate the entire CI build](https://github.com/arcus-azure/arcus/issues/292) as it would otherwise lead to too much test runs and unnecessary interaction problems.

For now, I would say: let go of the threshold.

stijnmoreels:
Waiting for https://github.com/arcus-azure/arcus/issues/292

PR #397, opened 2023-06-02
feat: use scoped message handlers via pumps/routers

### LINKED ISSUE #395: Limit scope of message handlers by only running registered handlers linked to message pump
Currently, message pumps are using message handlers registered from other pumps when processing messages due to the fact that all the instances are registered in the same application services container.

We have recently added a `JobId` property to the message handler collection type that registers the message handlers. This ID can be used to within the message handler registration so that when it is retrieved from a wrong message pump, it will respond with a 'can not process this message' result.

**Additional context**
Introduction of `JobId` in message handler collection: #333

stijnmoreels:
Enhances the current message handler registration by scoping any message handler registrations to a pump or router. This means that message handlers registered for another pump/router cannot be used during message routing.

This is a follow-up on the background jobs issue: https://github.com/arcus-azure/arcus.backgroundjobs/issues/200

The cool 😎 thing about this enhancement, is that it does not changes anything for single-message pump/router registrations; but changes the way we work with multiple pumps (which is the case when background jobs are involved).

Closes #395

### DIFF
--- a/docs/preview/02-Features/04-service-bus-extensions.md
+++ b/docs/preview/02-Features/04-service-bus-extensions.md
@@ -168,6 +168,10 @@ This information can be access in a more simplified way:
 
 ```csharp
 using Arcus.Messaging.Abstractions;
+using Azure.Messaging.ServiceBus;
+
+// Creates a new messaging context from the message, using an unique job ID to identify all message handlers that can handle this specific context.
+AzureServiceBusMessageContext messageContext = message.GetMessageContext("my-job-id");
 
 // Extract the encoding information from the `.ApplicationProperties` and wrapped inside a valid `Encoding` type.
 MessageContext messageContext = ...
--- a/docs/preview/02-Features/05-event-hubs-extensions.md
+++ b/docs/preview/02-Features/05-event-hubs-extensions.md
@@ -166,12 +166,13 @@ This information can be access in a more simplified way:
 
 ```csharp
 using Arcus.Messaging.Abstractions;
+using Azure.Messaging.EventHubs;
 
 EventProcessorClient processor = ... // Client that receives the message.
 EventData message = ... // The received message.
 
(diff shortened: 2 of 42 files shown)


stijnmoreels:
> > This means that message handlers registered for another pump/router cannot be used during message routing.
> 
> What do you mean exactly ?

In situations where you have two different message pumps/routers, the message handlers were previously shared among them. This PR makes sure that each set of message handlers is scoped to the pump/router; meaning that a message handler with the same message type will not 'wrongly' process a message that was meant for another pump/router.
This is especially the case with background jobs, where you process 'events'; and pumps have the same event/message type to process.

fgheysels:
> > > This means that message handlers registered for another pump/router cannot be used during message routing.
> > 
> > 
> > What do you mean exactly ?
> 
> In situations where you have two different message pumps/routers, the message handlers were previously shared among them. This PR makes sure that each set of message handlers is scoped to the pump/router; meaning that a message handler with the same message type will not 'wrongly' process a message that was meant for another pump/router. This is especially the case with background jobs, where you process 'events'; and pumps have the same event/message type to process.

Ok thanks.  But I would assume that this is something that could not ever happen, as the message handler is registered with a message-pump anyway ?

fgheysels (approved):

stijnmoreels:
> Ok thanks. But I would assume that this is something that could not ever happen, as the message handler is registered with a message-pump anyway ?

It seems that we due to the code syntax, but behind the scenes they are all registered in the same dependency container.
I succeeded in reproducing it by registering our two background jobs (auto-update secret in cache, auto-restart message pump), as they are both registered on 'secret new version created' events on Key Vault.
This PR makes sure that your assumption that the code syntax proposes is reality behind the scenes.

fgheysels (approved):

PR #407, opened 2023-07-13
feat: add configurable operation name

### LINKED ISSUE #404: EventHubsMessageRouting (?) in Azure Functions logs all invocations with operationname 'Process'
When using Arcus.Messaging to process messages coming from eventhub in an Azure Function, the Function invocations in Application Insights are logged with an operation-name 'Process'.

Is it possible to modify / specify the operation-name that must be used ?

stijnmoreels:
Add the possibility to configure the operation name when tracking the request telemetry in a message router.

Closes #404


docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -433,6 +433,10 @@ public class Startup
                 // (default: W3C).
                 options.Correlation.Format = MessageCorrelationFormat.Hierarchical;
 
+                // The name of the operation used when tracking the request telemetry.
+                // (default: Process)
+                options.Correlation.OperationName = "Order";
```
fgheysels, line 438: I'm not sure if it's a good idea to put this under 'Correlation' ?  It is imho not really related to 'correlation'.

stijnmoreels: More telemetry than correlation, maybe?

stijnmoreels: Currently, we do not have any properties directly on the options type. They are all grouped in different categories (serialization, correlation, routing...). That way, additional options for a specific category remain grouped and are not scattered.

fgheysels: I see.  However, I've been looking in the codebase and correct me if I'm wrong, but there's currently no 'Telemetry' category as you've proposed ?

I think it's a good idea to group properties together per 'category'.  However, I have currently no idea on how to name a category that contains just some 'common' or 'global' properties.

I was curious how this was categorized in AppInsights , and saw that for 'dependencies' this is logged under 'Dependency properties', for requests, it is categorized under 'Request properties', etc...

I think I wouldn't mind to have some global properties in the 'root' of the `MessageRouterOptions` class.  On the other hand, I also understand that this might quickly clutter the class if more properties would be added there.
What about `ContextualOptions` or something similar ?

stijnmoreels: Yes, was planning to create a new category for this.

Yes, it is somewhat related to 'context' and 'telemetry'. Would be more a fan to placing it there, indeed, than the root, as we've already updated this options type quite a lot and each time we were able to group them together bc of this approach.

stijnmoreels: Moved to a `Telemetry` sub-section. 👍

fgheysels (approved):

fgheysels:
Can you check the merge conflicts @stijnmoreels  ?

PR #412, opened 2023-08-02
feat: add managed identity to az eventhubs message pump

### LINKED ISSUE #409: Subscribe on a messaging service (servicebus, eventhubs, ...) using Managed Identity
I'm not sure if this is already possible, but I haven't found anything on it in the docs.

AFAIK, we must always provide a connection-string to the message-pump in order to be able to connect to Service Bus, EventHub, etc...
It would be nice if we could achieve that using managed identity as well.   The App that is created must have an identity then of course, and that identity must be assigned to the correct roles in Azure for this to work.   
If this would be possible, this would be a huge security improvement imho.

(Maybe this is already possible without having to make a code change ?  It's a matter of passing in the URL to the endpoint which doesn't contain any tokens / secrets, and setting the correct roles on identity level).

stijnmoreels:
Add the possibility to use Azure Managed Identity authentication when registering an Azure EventHubs message pump.

Closes #409

### DIFF
--- a/docs/preview/02-Features/02-message-handling/03-event-hubs.md
+++ b/docs/preview/02-Features/02-message-handling/03-event-hubs.md
@@ -69,7 +69,7 @@ public class Program
         services.AddSecretStore(stores => ...);
 
         // Add Azure EventHubs message pump and use OrdersMessageHandler to process the messages.
-        services.AddEventHubsMessagePump("<my-eventhubs-name>", "Arcus_EventHubs_ConnectionString", "<my-eventhubs-blob-storage-container-name>", "Arcus_EventHubs_Blob_ConnectionString")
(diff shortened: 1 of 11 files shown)


docs/preview/02-Features/02-message-handling/03-event-hubs.md
```diff
@@ -69,7 +69,7 @@ public class Program
         services.AddSecretStore(stores => ...);
 
         // Add Azure EventHubs message pump and use OrdersMessageHandler to process the messages.
-        services.AddEventHubsMessagePump("<my-eventhubs-name>", "Arcus_EventHubs_ConnectionString", "<my-eventhubs-blob-storage-container-name>", "Arcus_EventHubs_Blob_ConnectionString")
+        services.AddEventHubsMessagePumpUsingManagedIdentity("<my-eventhubs-name>", "Arcus_EventHubs_ConnectionString", "<my-eventhubs-blob-storage-container-name>", "Arcus_EventHubs_Blob_ConnectionString")
```
stijnmoreels, line 72: Also used the managed identity option as a default here, to guide the recommended way.

fgheysels (approved):

Issue #411, opened 2023-07-19
Support deferred messages or delayed processing


fgheysels:
We might have processes where we receive a message from servicebus that cannot be processed yet.  (For instance because we're waiting on another message coming from another party first).
In such situations, it would be good if we could defer the message or resubmit the same message.
It would be nice if we could do this in the implementation of the message-handler.  I can see that the `AzureServiceBusMessageHandler` now already has functionality to Complete or Abandon messages via protected methdos.  Some 'defer' or 'resubmit' logic there would be appropriate I think.

I think there are 2 approaches:

## Re-submitting:
I'm thinking of having the possibility of re-submitting the exact same message to Service Bus again, but with a specific / specified 
 `ScheduledEnqueueTimeUtc` property. [Docs on this here](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.message.scheduledenqueuetimeutc?view=azure-dotnet).  (That would probably mean completing the initial message and then resubmitting a copy of it , with all the same properties etc... but with a specified ScheduledEnqueueTimeUtc ?
(Or can this be implemented by abandoning the message and passing in the necessary 'propertiesToModify' ?  [ pass in `ScheduledEnqueueTimeUtc` as propertyToModify ? )

## Deferring

This might be another possible approach, but I think it's more complex?
Some documentation on the defer-functionality that is provided by Azure ServiceBus can be found [here](https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-deferral).
The drawback of deferring -as I see it-, is that you need another mechanism to retrieve deferred messages again, and that you need to specify which deferred message you want to retrieve again (so you need to keep state / administration for this).

stijnmoreels:
That would be a cool feature, indeed. I see some similarity with #25 and #36 .

fgheysels:
> That would be a cool feature, indeed. I see some similarity with #25 and #36 .

I see some similarity with #25 indeed.  However, @tomkerkhove mentions there that the pump should orchestrate this, but I don't know how the pump could do that.  The pump has no knowledge of the 'business logic' that must be executed (what must be executed first).
Therefore, my assumption would be that the message-handler would be responsible for 'resubmitting' the message if it cannot be processed yet.

fgheysels:
I've tried implementing this by calling the `AbandonAsync` method and specify a `scheduledEnqueueTimeUtc` property while abandoning, but this doesn't give us the desired results.

[...]

PR #417, opened 2023-09-22
chore: use built-in service bus client in message pump (🚧)

stijnmoreels:
Use the default registration of the `ServiceBusClient` to use in the Azure Service Bus message pump. This is somewhat related to the #411 to get a `ServiceBusSender` in a more easier way; but is also a performance improvement as the host will now handle the connection for us instead of the message pump (which was re-creating the client in some cases of restart).

Relates to #411


fgheysels (commented):
Why still in draft ?

stijnmoreels:
> Why still in draft ?

Bc I'm still working on it, as the tests were failing and it was late so I hadn't time to look into them. 😄 But I wanted to push my changes nonetheless so I would be reminded that I still need to work on this. So, thx! 😉

fgheysels:
Is this PR still relevant ?

PR #435, opened 2024-02-09
feat: do not automatically try to create and delete topic subscriptions

### LINKED ISSUE #432: Message pump tries to delete subscription on stop
It looks like the default-behavior of an Arcus MessagePump which is configured to consume a topic, is to delete the subscription it is using when the MessagePump stops:

image
Luckily my service principal is not allowed to delete the topic-subscription.

In my opinion, this is something that we should not do by default.  Nor should it be configured to create the subscription by default.

In the projects we're working on, we typically deploy all (Azure) resources via 'infrastructure as code'.  That means that we also setup the topic-subscriptions via IaC.   
A subscription also has filters; this is something that you carefully define and you define that upfront.

Next to that, suppose the message-pump runs in a Kubernetes cluster or Container App environment.   It is very possible that the component suddenly restarts for some reason and then restarts; in such cases, you really don't want that the subscription is being deleted when the message-pump stops, and is re-created automatically when the pump restarts (the filters that have been defined will not be automatically correctly recreated).

fgheysels:
Change the default behaviour of the `AzureServiceBusMessagePump`.
Instead of automatically creating a subscription on startup by default, and removing the subscription again on shut-down, we no longer do that.

Closes #432

### DIFF
--- a/docs/preview/02-Features/02-message-handling/01-service-bus.md
+++ b/docs/preview/02-Features/02-message-handling/01-service-bus.md
@@ -462,7 +462,7 @@ public class Startup
                 options.Deserialization.AdditionalMembers = AdditionalMemberHandling.Ignore;
 
                 // Indicate whether or not a new Azure Service Bus Topic subscription should be created/deleted
-                // when the message pump starts/stops (default: CreateOnStart & DeleteOnStop).
+                // when the message pump starts/stops (default: None, so no subscription will be created or deleted).
                 options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
             });
 
--- a/docs/preview/02-Features/02-message-handling/01-service-bus.md
+++ b/docs/preview/02-Features/02-message-handling/01-service-bus.md
@@ -463,7 +463,7 @@ public class Startup
 
                 // Indicate whether or not a new Azure Service Bus Topic subscription should be created/deleted
                 // when the message pump starts/stops (default: None, so no subscription will be created or deleted).
-                options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
+                options.TopicSubscription = TopicSubscription.Automatic;
             });
 
         services.AddServiceBusQueueMessagePump(
--- a/src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
+++ b/src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
@@ -36,7 +36,7 @@ namespace Arcus.Messaging.Tests.Unit.ServiceBus
                     "topic name", "subscription name", "secret name", options =>
                     {
                         options.AutoComplete = true;
-                        options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
+                        options.TopicSubscription = TopicSubscription.Automatic;
                     });
             
             // Assert
@@ -91,7 +91,7 @@ namespace Arcus.Messaging.Tests.Unit.ServiceBus
                     "topic name", "subscription name", "secret name", configureMessagePump: options =>
                     {
                         options.AutoComplete = true;
-                        options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
+                        options.TopicSubscription = TopicSubscription.Automatic;
                     });
 
             // Assert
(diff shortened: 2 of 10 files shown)


src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
```diff
@@ -33,7 +33,11 @@ public async Task AddServiceBusTopicMessagePump_WithSubscriptionNameIndirectSecr
             // Act
             ServiceBusMessageHandlerCollection result = 
                 services.AddServiceBusTopicMessagePump(
-                    "topic name", "subscription name", "secret name", options => options.AutoComplete = true);
+                    "topic name", "subscription name", "secret name", options =>
+                    {
+                        options.AutoComplete = true;
+                        options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
```
fgheysels, line 39: I'm actually wondering if it is necessary to have these 2 possibilities separated.
I think the pump should only be able to remove the subscription if the subscription has been auto-created.
In other words, I think it is useless to be able to only enable the `DeleteOnStop` option.  
I think we should combine these 2 flags into one setting:  `CreateOnStartAndDeleteOnStop` or `CreateAndDeleteSubscription` , something like that.

stijnmoreels: Yes, makes kinda sense, I believe. `CreateOnStart_DeleteOnStop`, `Temporary`?

stijnmoreels: We could introduce a new enumeration field with this new name, and deprecate the other two, as we haven't created a release with a warning of their deletion.

stijnmoreels:
Propably, some integration tests will have to change, too, for this to work. As the tests will assume that their test generated subscription is auto-created

src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
fgheysels: `AutoCreateAndDelete` ?
I can work on this later this week.

fgheysels: `AutoSetupSubscription`
`OwnsSubscription`
...

stijnmoreels: Maybe `Automatic` will already be enough, as the enumeration already is called `TopicSubscription`?

fgheysels: Ok.  
But I'm also thinking:  we have an enum now with only 2 options: 'None' and 'Automatic'.
Maybe this should be converted to a boolean flag ?

stijnmoreels: It would take more effort in adding any additional option we're not thinking of, then to go from `bool` to enum again, right? Just thinking library develeopment, here. 😀

fgheysels: That's true.  Just wondering what additional options will be added ....   But ok, let's stick with enum.

docs/preview/02-Features/02-message-handling/01-service-bus.md
```diff
@@ -462,7 +462,7 @@ public class Startup
                 options.Deserialization.AdditionalMembers = AdditionalMemberHandling.Ignore;
 
                 // Indicate whether or not a new Azure Service Bus Topic subscription should be created/deleted
-                // when the message pump starts/stops (default: CreateOnStart & DeleteOnStop).
+                // when the message pump starts/stops (default: None, so no subscription will be created or deleted).
                 options.TopicSubscription = TopicSubscription.CreateOnStart | TopicSubscription.DeleteOnStop;
```
stijnmoreels, line 466:
```suggestion
                options.TopicSubscription = TopicSubscription.Automatic;
```

src/Arcus.Messaging.Tests.Unit/ServiceBus/IServiceCollectionExtensionsTests.cs
stijnmoreels: Will resolve this as we stick with enum. 👍

PR #441, opened 2024-05-06
chore: all package upgrades FE docs

laurensvc:


docs/pnpm-lock.yaml
```diff
@@ -0,0 +1,10284 @@
+lockfileVersion: '9.0'
```
stijnmoreels, line 1: Could you also maybe give us some context as to why we have this lock file besides the package lock file?

stijnmoreels (approved):

docs/pnpm-lock.yaml
laurensvc: I am using a new package manager
PNPM instead of NPM
this makes upgrading way easier (for me)
normally this could result in some issues, since it can change some versions of packages you don't intent.
But since I am upgrading all packages anyway, this issue won't be present. :)

stijnmoreels: Ok, great, so could older package files then be deleted?

fgheysels (approved):

PR #442, opened 2024-06-20
fix: only fail on fallback when az servicebus message wasn't tried

### LINKED ISSUE #424: Not using a Fallbackhandler should be a valid use case and not throw an exception.
In our implementation of arcus messaging we do not want to use a fallback handler if an exception is thrown in the specified handler for the message. We now see 2 exceptions about a missing fallback handler that are covering up the initial exception thrown in the handler.
Which is polluting our application insights.

Not throwing an exception on missing fallback handler if not using a fallback handler is a choice.

Maybe make the exception of a missing fallback handler configurable and log a debug / trace message instead.

stijnmoreels:
Only when an incoming Azure Service bus message hasn't gone trhough a matching but failing message handler, should we throw if there were no fallback message handlers registered.

Closes #424


fgheysels (approved):
Great!  I like the approach of returning as soon as the message was processed.

PR #444, opened 2024-06-27
fix: use message receiver io processor to correctly stop message pump during circuit breaker

stijnmoreels:
The `ServiceBusProcessor` was not able to be stopped within its message processing - this PR uses the `ServiceBusReceiver` fully with a more recent Service Bus Messaging package.


docs/preview/02-Features/06-general-messaging.md
```diff
@@ -36,9 +36,16 @@ public class OrderMessageHandler : IAzureServiceBusMessageHandler<Order>
...
-        await _circuitBreaker.PauseMessageProcessingAsync(messageContext.JobId);
+        if (!IsDependentSystemHealthy())
+        {
+            // If not, call the circuit breaker, processing will be halted temporarily.
+            await _circuitBreaker.PauseMessageProcessingAsync(messageContext.JobId);
+        }
+        else
+        {
+            // If the dependent system is healthy, mark the circuit as closed.
+            await _circuitBreaker.ResumeMessageProcessingAsync(messageContext.JobId);
```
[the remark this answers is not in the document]
stijnmoreels: No, means that I forgot to update the documenttion. 😀

docs/preview/02-Features/06-general-messaging.md
```diff
@@ -47,8 +54,8 @@ The message pump will by default act in the following pattern:
 * Circuit breaker calls `Pause`
   * Message pump stops processing messages for a period of time (circuit is OPEN).
 * Message pump tries processing a single message (circuit is HALF-OPEN).
-  * Dependency still unhealthy? => Tries again later (circuit is OPEN)
-  * Dependency healthy? => Message pump starts receiving message in full again (circuit is CLOSED).
+  * Dependency still unhealthy? => circuit breaker pauses again (circuit is OPEN)
+  * Dependency healthy? => circuit breaker resumes, message pump starts receiving message in full again (circuit is CLOSED).
```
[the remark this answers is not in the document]
stijnmoreels: Yeah, forgot to update the documentation with the new setup.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -44,7 +65,12 @@ protected MessagePump(IConfiguration configuration, IServiceProvider serviceProv
         /// <summary>
         /// Gets the boolean flag that indicates whether the message pump is started and receiving messages.
         /// </summary>
-        public bool IsStarted { get; private set; }
+        public bool IsStarted { get; protected set; }
+
+        /// <summary>
+        /// Gets the current state of the message pump within the circuit breaker context.
+        /// </summary>
+        protected MessagePumpCircuitState CircuitState { get; set; } = MessagePumpCircuitState.Closed;
```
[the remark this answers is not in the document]
stijnmoreels: The service bus message pump sets this, but we might be able to make this private indeed.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -213,17 +212,13 @@ public override async Task<MessageProcessingResult> TryProcessProcessSingleMessa
...
+            CircuitState = MessagePumpCircuitState.HalfOpen;
             Logger.LogDebug("Azure Service Bus {EntityType} message pump '{JobId}' on entity path '{EntityPath}' in namespace '{Namespace}' tries to process single message during half-open circuit...", Settings.ServiceBusEntity, JobId, EntityPath, Namespace);
 
-            ServiceBusReceivedMessage message = null;
-            while (message is null)
-            {
-                message = await _messageReceiver.ReceiveMessageAsync();
-            }
-
+            ServiceBusReceivedMessage message = await _messageReceiver.ReceiveMessageAsync(maxWaitTime: TimeSpan.MaxValue);
```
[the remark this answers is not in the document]
stijnmoreels: The thing is, it returns `null` if no wait time has been specified, and in this half-open state, we should continue to be in the half-open state until there is a message available.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -242,24 +237,39 @@ public override async Task StartProcessingMessagesAsync(CancellationToken cancel
...
+            _receiveMessagesCancellation = new CancellationTokenSource();
+            while (CircuitState is MessagePumpCircuitState.Closed 
+                   && !_messageReceiver.IsClosed 
+                   && !_receiveMessagesCancellation.IsCancellationRequested)
+            {
+                try
+                {
+                    await foreach (ServiceBusReceivedMessage message in _messageReceiver.ReceiveMessagesAsync(_receiveMessagesCancellation.Token))
+                    {
+                        await ProcessMessageAsync(message, cancellationToken);
```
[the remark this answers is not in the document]
stijnmoreels: Is that even possible with the `IAsyncEnumerable`? Isn't it the purpose of that interface to force code constructs in this setup?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -44,7 +65,12 @@ protected MessagePump(IConfiguration configuration, IServiceProvider serviceProv
         /// <summary>
         /// Gets the boolean flag that indicates whether the message pump is started and receiving messages.
         /// </summary>
-        public bool IsStarted { get; private set; }
+        public bool IsStarted { get; protected set; }
```
fgheysels, line 68: The comment is a bit confusing.
The pump can be started, but not receiving messages because the CircuitState is open.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -125,6 +154,8 @@ public virtual Task StartProcessingMessagesAsync(CancellationToken cancellationT
         public virtual Task StopProcessingMessagesAsync(CancellationToken cancellationToken)
         {
             IsStarted = false;
+            CircuitState = MessagePumpCircuitState.Open;
```
fgheysels, line 157: Just wondering why stopping the pump should have an effect on the CircuitState ?

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
```diff
@@ -276,29 +248,15 @@ private string DetermineEntityPath(string connectionString = null)
...
         {
-            var messageHandlerOptions = new ServiceBusProcessorOptions();
+            var options = new ServiceBusReceiverOptions();
             if (Options != null)
             {
-                messageHandlerOptions.AutoCompleteMessages = Options.AutoComplete;
-                messageHandlerOptions.MaxConcurrentCalls = Options.MaxConcurrentCalls;
-                messageHandlerOptions.PrefetchCount = Options.PrefetchCount;
+                options.PrefetchCount = Options.PrefetchCount;
             }
```
fgheysels, line 257: Just wondering what we now do with MaxConcurrentCalls ?  This is quite an important setting but I don't immediately see from this PR where we are now using this ?  
This controls how many messages we should retrieve from the bus in one batch.

fgheysels: Ok, I've found it here in the `StartProcessingMessagesAsync` method:

```
IReadOnlyList<ServiceBusReceivedMessage> messages = 
                        await _messageReceiver.ReceiveMessagesAsync(Settings.Options.MaxConcurrentCalls, cancellationToken: _receiveMessagesCancellation.Token);
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -242,24 +239,39 @@ public override async Task StartProcessingMessagesAsync(CancellationToken cancel
...
+                catch (Exception exception) when (exception is TaskCanceledException or OperationCanceledException or ObjectDisposedException)
+                {
+                    IsStarted = false;
+
+                    Logger.LogTrace("Azure Service Bus {EntityType} message pump '{JobId}' on entity path '{EntityPath}' in '{Namespace}' was cancelled", Settings.ServiceBusEntity, JobId, EntityPath, Namespace);
+                    return;
+                }
+                catch (Exception exception)
+                {
+                    await ProcessErrorAsync(exception, cancellationToken);
```
fgheysels, line 272: This is a bit problematic I guess.
I've been playing around with this in a demo-project and I simulate an error that occurs during processing by throwing an exception.
Normally, the message that could not be processed must be re-delivered to the queue (or topic) so that it can be retried in a next iteration.   Eventually, when the message has been re-delivered a number of time, ServiceBus will deadletter the message.

So, in my demo-app I expect to see some messages in the deadletter queue, but there are none present.  This means that failed messages are being removed from the queue/topic because I believe the 'Complete' method is called as the exception that is thrown in the message-handler is eventually caught here and 'eaten'.
In case of an unhandled exception, we must call the Abandon functionality for the message that could not be processed.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: This is reusing existing functionlity. The state is automatically set whithin the lifetime methods. When the pump is stopped, than the state is always open.

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
stijnmoreels: Yes, indeed.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
stijnmoreels: Ah, abondoning was indeed something that the processor did for us, but we are now responsible with the receiver. Did not noitice. Will add/test this.

stijnmoreels: In case of 'exception swallowing', the messages are auto-abandonned when they does not seem to be processed by the receiver (delivery count is increased). That I got verified in the tests. I added the auto-dead-letter in the message receiver when no message handler was able to process the message.

docs/preview/02-Features/06-general-messaging.md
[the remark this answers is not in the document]
stijnmoreels: This is now changed in the docs.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: This is more closely linked to each other nd automatically assigned internally. The state is tracked by start/stoping the pump, just s this boolean. There is no 'cricuit breaker'-specific methods exposed, only the start/stop methods that internally sets the boolean and the state. The boolean here acts as a more friendly exposure of the 'state' of the pump.

fgheysels: I just wanted to say that it's maybe better to remove the part `and receiving messages` ?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
fgheysels: Related to this, I'm wondering if we even have a scenario where the `AutoComplete` option would be set to `false` ?
Is that something that we should continue to support ?  (Breaking change, and also outside the scope of this PR, but just something I was thinking about).

stijnmoreels: If you wanted to complete the message yourself with one of the message handlers (fallback or general), I would think. It gives a bit more control.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Yes, that we can do.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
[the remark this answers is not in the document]
fgheysels: If it should only be set via a set of methods, I'd make this private.
What would happen if an inheritor sets this property directly ?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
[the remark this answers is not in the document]
fgheysels: So, we're just indefintely waiting until a message is received.
OK then.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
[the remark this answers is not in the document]
stijnmoreels: The service bus message pump needs this, I believe, to activate it back again. I believe that was also the reason it was made `protected`, so that client implementations of the message pump have control. The `MessagePump` does not contain any functionality.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
fgheysels: > In case of 'exception swallowing', the messages are auto-abandonned when they does not seem to be processed by the receiver (delivery count is increased). That I got verified in the tests. I added the auto-dead-letter in the message receiver when no message handler was able to process the message.

That's strange, I've been playing a bit with the latest version (2.0.0-PR-444-20241007.5) and I intentionally throw an exception in the message-handler that is not being caught.  The messages end up in the deadletter queue, however, their Delivery Count property remains 0, which is not wat I expected

image

I expected that the DeliveryCount would have been set to 10 and ServiceBus should've assigned a deadletter reason saying that the max delivercount has been reached.

fgheysels: Did some further digging; I throw an exception in the MessageHandler which is not being caught.
I would expect that the servicebus message is abandoned and is being returned to the (in this case) queue .  That would cause that the message can be picked up again, and the message-handler would process the message again.
However, this doesn't happen.  The message-handler receives the message for the first time, I throw an exception and the message is being deadlettered right away, although my 'application code' doesn't contain any statement that would deadletter the message.

stijnmoreels: Yes that that is expected behavior, actually. We already talked about this - we can't know if a message handler throws an exception because it can't handle it, or because the backend system is failing, or the implementation is wrong... We only know that it can't match it, so we go to the next message handler and if there is none, then we dead-letter.

fgheysels: We should not deadletter immediately.  The message must be abandoned which means it's delivered back to the queue or topic.
The message will be received again, and will be routed once again through the message-handlers.   If we have a matching message-handler, it will process the message once again (you can see it as the message processing being retried).  And this goes on as many times until the message gets finally succesfully processed or until the max amount of deliveries has been reached.  In the latter case, the message will be deadlettered.
What I've described is also the behaviour that currently exists in Arcus.Messaging 2.0.0

stijnmoreels: But without the receiver, then. Hmm, okay, then we need to investigate a bit more on this on how to verify this - will be back on this Monday.

fgheysels: It looks like the deadlettering is now working as expected, however, the circuitbreaker doesn't seem to work anymore.  (I see that there are also some tests that are failing ? )

I have created a small test project that consumes messages from a queue and calls an API.  When the API is unavailable, the circuitbreaker is triggered.  
I can see that the messagepump stops (which is expected).  After a while, I can see the circuitbreaker transitioning into a half-open state, as a new message is received.    When the API is available and the message can be processed as expected, I'd expect that the circuitbreaker goes into the closed state, and the messagepump starts retrieving messages from the queue at max configured speed.  However, this doesn't happen.  Moreover, it seems that the MessagePump isn't retrieving anything at all.  Messages remain in the queue.
Even if `ResumeMessageProcessingAsync` is called by the MessageHandler, nothing seems to happen.

There are some failing tests so that might be the indication that there's something going wrong ?

stijnmoreels: Ok, so some additional testing is required - this was actually done better in another PR where we already worked with the new Dev tenant, but we'll have to check here first, I guess.
Currently, not being able to simulate the failure as the tests show a message pump that received messages even after a circuit breaker hapenned. Will need some time to investigate how to come to the same situation.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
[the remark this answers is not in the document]
stijnmoreels: To clarify: the `MessagePump` does not have any functionality and acts here more as an interface than as the parent class. It provides the necessary properties and start/stop information, but not the actual implementation. What it does help us with, is streamlining message pumps and adding functionality to pumps without knowing its implementation.

The circuit breaker state here is only for the implementations to use, and in this case the half-open state is something that this interface does not know where to set.

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
```diff
@@ -16,5 +16,12 @@ public interface IMessagePumpCircuitBreaker
         /// <param name="configureOptions">The optional user-configurable options to manipulate the workings of the message pump interaction.</param>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
         Task PauseMessageProcessingAsync(string jobId, Action<MessagePumpCircuitBreakerOptions> configureOptions);
+
```
fgheysels, line 19: The interface should have a property that exposes the state of the CircuitBreaker (open, closed, halfopen).

src/Arcus.Messaging.Pumps.ServiceBus/Resiliency/CircuitBreakerServiceBusMessageHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+                await CircuitBreaker.PauseMessageProcessingAsync(messageContext.JobId, opt =>
+                {
+                    opt.MessageIntervalDuringRecovery = options.MessageIntervalDuringRecovery;
+                    opt.MessageRecoveryPeriod = options.MessageRecoveryPeriod;
+                });
+                await AbandonMessageAsync();
+                throw result.ProcessingException;
+            }
+
+            await CircuitBreaker.ResumeMessageProcessingAsync(messageContext.JobId);
```
fgheysels, line 68: We should not always just call ResumeMessageProcessing I think.
This should only be called when the circuit-breaker is in the half-open state.
(Suppose the circuitbreaker is just triggered, but you still have the remainder of the messages that have been retrieved by the pump that must be processed.  You don't want to close the circuitbreaker again because one of those messages can be processed.).

fgheysels:
I'm testing this a bit further and I'm getting good results; however, I now run into an issue where an exception was thrown for some reason (no idea what happened).
The exception was thrown, and after that exception, some other traces were logged saying that the received message was 'null'.
And afterwards, no new messages were read from ServiceBus by the MessagePump.

```
[23:02:26 ERR] Azure Service Bus Queue message pump 'a885a71c-74e8-4325-a681-e9fccbf1a9f6' on entity path 'orders' in namespace 'fg-sb-arcusdemo.servicebus.windows.net' failed to process single message during half-open circuit, retrying after circuit delay
Azure.Messaging.ServiceBus.ServiceBusException: The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue. For more information please see https://aka.ms/ServiceBusExceptions . Reference:4cc34e78-a2da-47ab-9963-c4e1e30659e8, TrackingId:f6a8ee520000004d01ba96c6672be710_G7_B18, SystemTracker:gi::G7:720044332:amqps://fg-sb-arcusdemo.servicebus.windows.net/-e1a11b51;0:5:6:source(address:/orders,filter:[]), bi::in-connection50454(G7-38509)::session50457::link29005510, Timestamp:2024-11-06T22:02:26 (MessageLockLost). For troubleshooting information, see https://aka.ms/azsdk/net/servicebus/exceptions/troubleshoot.
   at Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.DisposeMessageAsync(Guid lockToken, Outcome outcome, DispositionStatus disposition, TimeSpan timeout, IDictionary`2 propertiesToModify, String deadLetterReason, String deadLetterDescription)
   at Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.CompleteInternalAsync(Guid lockToken, TimeSpan timeout)
   at Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.<>c.<<CompleteAsync>b__47_0>d.MoveNext()
--- End of stack trace from previous location ---
   at Azure.Messaging.ServiceBus.ServiceBusRetryPolicy.<>c__22`1.<<RunOperation>b__22_0>d.MoveNext()
--- End of stack trace from previous location ---
   at Azure.Messaging.ServiceBus.ServiceBusRetryPolicy.RunOperation[T1,TResult](Func`4 operation, T1 t1, TransportConnectionScope scope, CancellationToken cancellationToken, Boolean logTimeoutRetriesAsVerbose)
   at Azure.Messaging.ServiceBus.ServiceBusRetryPolicy.RunOperation[T1,TResult](Func`4 operation, T1 t1, TransportConnectionScope scope, CancellationToken cancellationToken, Boolean logTimeoutRetriesAsVerbose)
   at Azure.Messaging.ServiceBus.ServiceBusRetryPolicy.RunOperation[T1](Func`4 operation, T1 t1, TransportConnectionScope scope, CancellationToken cancellationToken)
   at Azure.Messaging.ServiceBus.Amqp.AmqpReceiver.CompleteAsync(Guid lockToken, CancellationToken cancellationToken)
   at Azure.Messaging.ServiceBus.ServiceBusReceiver.CompleteMessageAsync(ServiceBusReceivedMessage message, CancellationToken cancellationToken)
   at Arcus.Messaging.Pumps.ServiceBus.AzureServiceBusMessagePump.ProcessMessageAsync(ServiceBusReceivedMessage message, CancellationToken cancellationToken)
   at Arcus.Messaging.Pumps.ServiceBus.AzureServiceBusMessagePump.TryProcessProcessSingleMessageAsync(MessagePumpCircuitBreakerOptions options)
 ```

File: src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs (hunk 10 above)
Comment (stijnmoreels) [reply src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs:26]: Done with the `GetMessageProcessingState` on the circuit breaker.

File: src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs [hunk 12]
```diff
@@ -9,12 +9,26 @@ namespace Arcus.Messaging.Pumps.Abstractions.Resiliency
     /// </summary>
     public interface IMessagePumpCircuitBreaker
     {
+        /// <summary>
+        /// Gets the current circuit breaker state of message processing in the given message pump.
+        /// </summary>
+        /// <param name="jobId">The unique identifier to distinguish the message pump in the application services.</param>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
+        MessagePumpCircuitState GetMessageProcessingState(string jobId);
```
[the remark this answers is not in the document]
stijnmoreels: Yeah, wanted to be a bit in line with the other 'message processing' methods. But yes, we can do that. 👍

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
```diff
@@ -44,25 +43,62 @@ public virtual async Task PauseMessageProcessingAsync(string jobId, Action<Messa
...
 
             MessagePump messagePump = GetRegisteredMessagePump(jobId);
             if (messagePump.IsStarted)
             {
+                _logger.LogDebug("Open circuit by pausing message processing for message pump '{JobId}'...", jobId);
                 await messagePump.StopProcessingMessagesAsync(CancellationToken.None);
-                await WaitRecoveryTimeAsync(messagePump, options);
-                await TryProcessSingleMessageAsync(messagePump, options);
+
+                await Task.Factory.StartNew(async () =>
```
fgheysels, line 55: I've been testing the CircuitBreaker functionality a bit more, and I have inconsistent results.
Sometimes the queue I'm processing gets completely processed, but more often then not, messages remain in the queue and the MessageProcessor is not retrieving anything anymore.

I suspect that this might have to do with these tasks that are started in the background.

Wouldn't it be an idea to change the 'strategy' of the message-receiver when the circuit-breaker transitions into a half-open state instead ?

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
```diff
@@ -95,33 +131,19 @@ protected MessagePump GetRegisteredMessagePump(string jobId)
...
-                    _logger.LogTrace("Message pump '{JobId}' successfully handled a single message, closing circuit...", messagePump.JobId);
-                    await messagePump.StartProcessingMessagesAsync(CancellationToken.None);
-                }
-                else
-                {
-                    await WaitMessageIntervalAsync(messagePump, options);
-                }
-            }
+            _logger.LogDebug("Try to process single message in message pump '{JobId}' (state: half-open)", messagePump.JobId);
+            await messagePump.TryProcessProcessSingleMessageAsync(options);
```
fgheysels, line 135: This `TryProcessSingleMessageAsync` method of the message-pump retuns an indication whether or not the processing succeeded, shouldn't we return that value here in this private method as well and then do something with it ?

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
fgheysels: Let's park this for a while.

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
```diff
@@ -44,25 +43,62 @@ public virtual async Task PauseMessageProcessingAsync(string jobId, Action<Messa
...
             {
+                _logger.LogDebug("Open circuit by pausing message processing for message pump '{JobId}'...", jobId);
                 await messagePump.StopProcessingMessagesAsync(CancellationToken.None);
-                await WaitRecoveryTimeAsync(messagePump, options);
-                await TryProcessSingleMessageAsync(messagePump, options);
+
+                await Task.Factory.StartNew(async () =>
+                {
+                    await WaitRecoveryTimeAsync(messagePump, options);
+                    await TryProcessSingleMessageAsync(messagePump, options);
```
fgheysels, line 58: We're trying to process a single message, but what should we do next ?
Shouldn't we continue trying if processing that single message fails ?
In other words, shouldn't these 2 lines (L57 & 58) be executed in a loop until the processing succeeded ?


/just thinking out loud.

fgheysels: And once succeeded, can't the message-pump transition to the Closed state by itself ?

fgheysels: I've tried this locally, and seem to have good results with it.  Additional benefit is that as an 'application programmer', I don't have to worry myself about resuming the pump / closing the circuitbreaker.

I changed this on my local version to this:

```
 if (messagePump.IsStarted)
 {
     _logger.LogInformation("Open circuit by pausing message processing for message pump '{JobId}'...", jobId);
     await messagePump.StopProcessingMessagesAsync(CancellationToken.None);

     await Task.Factory.StartNew(async () =>
     {
         _logger.LogInformation("Starting recovery-task");
         MessageProcessingResult result;
         do
         {
             await WaitRecoveryTimeAsync(messagePump, options);
             _logger.LogInformation("Done waiting in recovery");
             result = await TryProcessSingleMessageAsync(messagePump, options);
         } while (result.IsSuccessful == false);

         await ResumeMessageProcessingAsync(jobId); 
     }, TaskCreationOptions.LongRunning);
 }
 else
 {
     throw new InvalidOperationException("Cannot pause when not running");
 }
```

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -213,24 +214,29 @@ public override async Task<MessageProcessingResult> TryProcessProcessSingleMessa
...
             Logger.LogDebug("Azure Service Bus {EntityType} message pump '{JobId}' on entity path '{EntityPath}' in namespace '{Namespace}' tries to process single message during half-open circuit...", Settings.ServiceBusEntity, JobId, EntityPath, Namespace);
 
             ServiceBusReceivedMessage message = null;
             while (message is null)
             {
                 message = await _messageReceiver.ReceiveMessageAsync();
+                if (message is null)
+                {
+                    Logger.LogTrace("Azure Service Bus {EntityType} message pump '{JobId}' on entity path '{EntityPath}' in namespace '{Namespace}' failed to receive a single message, trying again...", Settings.ServiceBusEntity, JobId, EntityPath, Namespace);
+                }
```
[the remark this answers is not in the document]
fgheysels: Added a delay in this commit https://github.com/arcus-azure/arcus.messaging/pull/444/commits/23520db71ecf7a185ff911e967515de245d1656c

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
fgheysels: Done here https://github.com/arcus-azure/arcus.messaging/pull/444/commits/23520db71ecf7a185ff911e967515de245d1656c

src/Arcus.Messaging.Pumps.ServiceBus/Resiliency/CircuitBreakerServiceBusMessageHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+            AzureServiceBusMessageContext messageContext,
+            MessageCorrelationInfo correlationInfo,
+            CancellationToken cancellationToken)
+        {
+            var options = new MessagePumpCircuitBreakerOptions();
+            MessageProcessingResult result = await TryProcessMessageAsync(message, messageContext, correlationInfo, options, cancellationToken);
+
+            if (!result.IsSuccessful)
+            {
+                await CircuitBreaker.PauseMessageProcessingAsync(messageContext.JobId, opt =>
```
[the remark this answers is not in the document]
fgheysels: This can actually be done when the developer doesn't use this messagehandler.

src/Arcus.Messaging.Pumps.ServiceBus/Resiliency/CircuitBreakerServiceBusMessageHandler.cs
fgheysels: This is no longer relevant.  The application developer is not responsible for calling resume himself with the changes we've made now.  
The MessagePump will automatically transition into a half-open and later in a closed state.

fgheysels (approved):
It was a long journey, and a lot of work has been done to make this work.
Thanks for your perseverance, I think we can now merge these changes :)

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
```diff
@@ -44,25 +43,62 @@ public virtual async Task PauseMessageProcessingAsync(string jobId, Action<Messa
         {
```
fgheysels, line 42: I don't understand - the code that we have here is completely different as opposed to what we had here:
https://github.com/arcus-azure/arcus.messaging/pull/453/files#diff-397dfb045522acf325b8c768be9bf7b7b69846d6118ce9bba9f57c96879b3de0

fgheysels: When I pull the repo and look at it locally, the file looks as expected though.

stijnmoreels: It's the same now, no? Think it's correct?

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
```diff
@@ -44,25 +43,62 @@ public virtual async Task PauseMessageProcessingAsync(string jobId, Action<Messa
...
+                }, TaskCreationOptions.LongRunning);
+                
             }
             else
             {
-                await WaitMessageIntervalAsync(messagePump, options);
-                await TryProcessSingleMessageAsync(messagePump, options);
+                await Task.Factory.StartNew(async () =>
+                {
+                    await WaitMessageIntervalAsync(messagePump, options);
```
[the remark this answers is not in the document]
stijnmoreels: Think this is now irrelevant as it has changed?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Removed the part here: https://github.com/arcus-azure/arcus.messaging/pull/444/commits/d5cfabf7798c51101c6dd1bfdcd44a7db74510dc

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
fgheysels: Yes it's OK.

fgheysels (approved):

PR #446, opened 2024-07-18
test: remove event-grid reference from service-bus integration tests

stijnmoreels:
Clean-up the Service Bus integration tests w/o any EventGrid reference for event sourcing, but use the local disk as the even source.

Also split the tests by category and clean-up duplications.


fgheysels:
there are some warnings in this PR regarding obsolete types that are still being used.  Should that be tackled as well in this PR ?

fgheysels (approved):

PR #453, opened 2024-11-26
feat: change circuitbreaker approach

fgheysels:
The CircuitBreaker now just signals the MessagePump to pause retrieving messages instead of stopping the MessagePump and restarting it later using a separate Task.


src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
```diff
@@ -443,6 +457,8 @@ protected async Task TryServiceBusFallbackMessageAsync(
                 Logger.LogWarning("No fallback message handler processed the Azure Service Bus message '{MessageId}' in pump '{JobId}', abandoning message!", message.MessageId, messageContext.JobId);
                 await messageReceiver.AbandonMessageAsync(message);
```
stijnmoreels, line 458: Wondering if this is correct?
Message could be either gone through a general message handler or not; no fallback is matched, but we still abandon the message instead of dead-lettering it?

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
```diff
@@ -40,70 +38,31 @@ public DefaultMessagePumpCircuitBreaker(IServiceProvider serviceProvider, ILogge
...
-            }
+            var options = new MessagePumpCircuitBreakerOptions();
+            configureOptions?.Invoke(options);
 
-            _logger.LogInformation("Message pump '{JobId}' successfully handled a single message, resume message processing (circuit breaker: closed)", messagePump.JobId);
-            await messagePump.StartProcessingMessagesAsync(CancellationToken.None);
+            _logger.LogDebug("Open circuit by pausing message processing for message pump '{JobId}'...", jobId);
 
+            messagePump.NotifyPauseReceiveMessages(options);
+            return Task.CompletedTask;
```
stijnmoreels, line 65: I would still leave the `PauseMessageProcessingAsync` asynchronous as the `Default...` one is only our default implementation of an interfact, not the only possible one.

src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePump.ResiliencyTests.cs
```diff
@@ -136,90 +138,6 @@ private async Task<TemporaryTopicSubscription> CreateTopicSubscriptionForMessage
                 rule: new CreateRuleOptions("MessageId", new SqlRuleFilter($"sys.messageid in ({string.Join(", ", messages.Select(m => $"'{m.MessageId}'"))})")));
         }
 
-        [Fact]
-        public async Task ServiceBusTopicMessagePump_PauseViaCircuitBreaker_RestartsAgainWithOneMessage()
```
stijnmoreels, line 140: Removed this test as it was mostly testing the wait time, not the circuit breaker functionality. Was useful during development, but unnecessary for the test suite as we have the `ServiceBusMessageQueuePump_WithUnavailableDependencySystem_CircuitBreaksUntilDependencyBecomesAvailable` and like tests that more check the circuit breaker functionality as a whole.

src/Arcus.Messaging.Pumps.ServiceBus/Resiliency/CircuitBreakerServiceBusMessageHandler.cs
```diff
@@ -61,11 +61,9 @@ await CircuitBreaker.PauseMessageProcessingAsync(messageContext.JobId, opt =>
                     opt.MessageIntervalDuringRecovery = options.MessageIntervalDuringRecovery;
                     opt.MessageRecoveryPeriod = options.MessageRecoveryPeriod;
                 });
-                await AbandonMessageAsync();
```
stijnmoreels, line 64: Taken this out as the test showed that sometimes messages get abandonned twice in the same receive operation - one time here, and one time by the message router.

src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePump.ResiliencyTests.cs
fgheysels: I think this test is valuable because it shows / verifies that the configured wait-times are respected ?

stijnmoreels: Yes, it shows that if we set a wait time, the system correctly uses the wait time: it is a complex get/set integration test that verifies little and has a bit setup cost. And what's more: the implementation can easily circumvent the assertion here by waiting on another point.
It was useful in the beginning as I used it during the initial test-driven development, but now, it feels out of touch with the circuit breaker functionality.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageProcessingResult.cs
```diff
@@ -1,24 +1,49 @@
...
+    /// <summary>
+    /// Represents all the possible errors of a <see cref="MessageProcessingResult"/>.
+    /// </summary>
+    public enum MessageProcessingError
+    {
+        /// <summary>
+        /// Defines an error that shows that the message processing was interrupted by some external cause,
+        /// unrelated to the message routing.
+        /// </summary>
+        ProcessingInterrupted,
```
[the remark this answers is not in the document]
stijnmoreels: We will run into the `ProcessingInterrupted` error when:
* The message router has an unexpected exception during message routing. We don't throw exceptions anymore, but there was an catch block that abandons the message in case of such unexpected exception. Maybe we should remove this catch block, as I believe that this is not expected behavior.
* Same goes for the message pump, there was a catch block arround the single message processing. Maybe this can also be removed.
* Then we have two references in the message pump where we `null` check a message and when the pump is cancelled.

> Exception thrown by messagehandler vs messagehandler returning 'no success' ?

This is the same thing. The message handler does not return anything, but if it throws, the `MatchedHandlerFailed` error will be used.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -154,11 +120,56 @@ public virtual Task StartProcessingMessagesAsync(CancellationToken cancellationT
...
+        /// <summary>
+        /// Waits a previously configured amount of time until the message pump is expected to be recovered (Closed to Open state).
+        /// </summary>
+        /// <param name="cancellationToken">The token to cancel the wait period.</param>
+        protected async Task WaitMessageRecoveryPeriodAsync(CancellationToken cancellationToken)
+        {
+            Logger.LogDebug("Circuit breaker caused message pump '{JobId}' to wait message interval of '{Interval}' during '{State}' state", JobId, CircuitState.Options.MessageRecoveryPeriod.ToString("g"), CircuitState);
+            await Task.Delay(CircuitState.Options.MessageRecoveryPeriod, cancellationToken);
+
+            CircuitState = CircuitState.TransitionTo(CircuitBreakerState.HalfOpen);
```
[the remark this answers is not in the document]
stijnmoreels: Now, we were in an open state, then the `WaitRecovery` was called by the Azure Service bus message pump, and after the wait, we're transitioning into the half-open state to see if a single message can be retrieved again.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -154,11 +120,56 @@ public virtual Task StartProcessingMessagesAsync(CancellationToken cancellationT
...
+        }
+
+        /// <summary>
+        /// Waits a previously configured amount of time until the next single message can be tried (Half-Open state).
+        /// </summary>
+        /// <param name="cancellationToken">The token to cancel the wait period.</param>
+        protected async Task WaitMessageIntervalPeriodAsync(CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
stijnmoreels: Wouldn't this obscure the two wait methods?
Now there is a clear difference between 'recovery' and 'interval', like the circuit breker options define.

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
```diff
@@ -23,12 +23,83 @@ public interface IMessagePumpCircuitBreaker
...
+        /// </summary>
+        internal static MessagePumpCircuitState Closed => new(CircuitBreakerState.Closed);
+
+        /// <summary>
+        /// Creates an instance of the <see cref="MessagePumpCircuitState"/> class that represents an open state,
+        /// in which the message pump is unable to process messages.
+        /// </summary>
+        /// <param name="options">The additional options to configure the message pump during the open state.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="options"/> is <c>null</c>.</exception>
+        internal static MessagePumpCircuitState Open(MessagePumpCircuitBreakerOptions options)
```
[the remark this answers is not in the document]
stijnmoreels: Well, the `Open` state introduces a new circuit-breaker context; which includes new options. The transition methods are for when already have the options received.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
```diff
@@ -407,19 +423,19 @@ private async Task DeleteTopicSubscriptionAsync(CancellationToken cancellationTo
...
         }
 
-        private async Task ProcessMessageAsync(ServiceBusReceivedMessage message, CancellationToken cancellationToken)
+        private async Task<MessageProcessingResult> ProcessMessageAsync(ServiceBusReceivedMessage message, CancellationToken cancellationToken)
         {
             if (message is null)
             {
                 Logger.LogWarning("Received message on Azure Service Bus {EntityType} message pump '{JobId}' was null, skipping", Settings.ServiceBusEntity, JobId);
-                return;
+                return MessageProcessingResult.Failure("<unavailable>", MessageProcessingError.ProcessingInterrupted, "Cannot process received message as the message is was 'null'");
```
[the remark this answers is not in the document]
stijnmoreels: Also not sure, to be honest.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -154,11 +120,56 @@ public virtual Task StartProcessingMessagesAsync(CancellationToken cancellationT
...
+
+        /// <summary>
+        /// Notifies the message pump about the new state which resumes message retrieval.
+        /// </summary>
+        protected void NotifyResumeRetrievingMessages()
+        {
+            Logger.LogDebug("Circuit breaker caused message pump '{JobId}' to transition back from '{CurrentState}' to a 'Closed' state, retrieving messages is resumed", JobId, CircuitState);
+
+            CircuitState = MessagePumpCircuitState.Closed;
+        }
```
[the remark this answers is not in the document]
stijnmoreels: Yes, and only the 'wait' ones and 'resume' are available for the users of custom message pump, or our Azure Service bus message pump.

These methods are shorthand methods that remove the need for implementations to know when the state transitions. The implementation only knows that it has to wait, or that it has to resume.

The 'notify' methods are for state switches. It is specifically called 'notify' instead of `Resume/Pause...` as it doesn't actually do those things, only, 'signals' such an action. Hence the name.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -154,11 +120,56 @@ public virtual Task StartProcessingMessagesAsync(CancellationToken cancellationT
...
+            CircuitState = CircuitState.TransitionTo(CircuitBreakerState.HalfOpen);
+        }
+
+        /// <summary>
+        /// Waits a previously configured amount of time until the next single message can be tried (Half-Open state).
+        /// </summary>
+        /// <param name="cancellationToken">The token to cancel the wait period.</param>
+        protected async Task WaitMessageIntervalPeriodAsync(CancellationToken cancellationToken)
+        {
+            Logger.LogDebug("Circuit breaker caused message pump '{JobId}' to transition into an '{State}' state, retrieving messages is paused for '{Period}'", JobId, CircuitState, CircuitState.Options.MessageIntervalDuringRecovery.ToString("g"));
```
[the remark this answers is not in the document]
stijnmoreels: Ah, yes, changed the log message here: https://github.com/arcus-azure/arcus.messaging/pull/453/commits/21e0b19ea77165df456ec3e6e418f165c191f22c

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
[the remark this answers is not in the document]
stijnmoreels: Yes, changed it here: https://github.com/arcus-azure/arcus.messaging/pull/453/commits/21e0b19ea77165df456ec3e6e418f165c191f22c

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
```diff
@@ -23,12 +23,83 @@ public interface IMessagePumpCircuitBreaker
...
-        /// <param name="jobId">The unique identifier to distinguish the message pump in the application services.</param>
-        /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
-        Task ResumeMessageProcessingAsync(string jobId);
+        public bool IsClosed => _state is CircuitBreakerState.Closed;
+
+        /// <summary>
+        /// Gets the boolean flag that indicates whether the message pump circuit breaker state is in the 'Open' state,
+        /// and should stop retrieving messages.
+        /// </summary>
+        public bool IsOpen => _state is CircuitBreakerState.Open;
```
fgheysels, line 64: I'm wondering why we don't expose the `CircuitBreakerState` itself (the value of the `_state` member) instead of introducing these 2 boolean properties ?
Don't we want to expose that the circuitbreaker can be in a half-open state ?

Being in a 'half open' state is of course an 'internal' thing.  IDK, just wondering what your idea / reasoning is here.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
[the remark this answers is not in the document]
fgheysels: I think it would be good if we could have some kind of 'extension point' here, so that another component can be informed when the CircuitBreaker changes into another state.
For instance, in my demo / poc application, I would like to be informed when the circuitbreaker transitions from half-open to closed state.
Maybe we can introduce some kind of event that is triggered whenever the circuitbreaker transitions into another state ?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs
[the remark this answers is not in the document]
fgheysels: We can leave it in here.  I wouldn't want to break anything :)

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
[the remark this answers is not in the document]
stijnmoreels: Ah, yes, that would be nice! Maybe some interface that we expose and that you can register in the depenceny container as 'event subscribers'? Or something else? You're the messaging expert 🙂.

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
stijnmoreels: Ah, it started as an usability/readability thing. As if we would expose the `_state` itself, you would have to expose the enumeration also and have conditions like these:

```csharp
if (CircuitState.State is CircuitBreakerState.Open)
```

While it was a bit more readable hide the enumeration and the state field and only expose those two start/stop states indeed.

```csharp
if (CircuitState.IsOpen)
```

src/Arcus.Messaging.Abstractions/MessageHandling/MessageProcessingResult.cs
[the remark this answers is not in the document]
fgheysels: - An unexpected exception during message routing is something else then a message for which the MessageRouter can't find a MessageHandler ?

- The catch block around the single message processing: does this catch the exceptions that might be thrown by the MessageHandler's implementation  ?  It seems that this is already handled somewhere else.   If so, I think we can indeed remove that catch block.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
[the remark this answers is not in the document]
fgheysels: I was initially more thinking on exposing an `event` in the CircuitBreaker, but that might be difficult to use, as you need to 'get' the CircuitBreaker out of the DI container, and subscribe to that event.  
The approach that you suggest sounds better.

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
fgheysels: Ok I see and understand.
However, when the CircuitBreaker is in a half-open state, both IsOpen & IsClosed will return false.  How do we cope with that ?

stijnmoreels: Will create `IsHalfOpen`.

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
stijnmoreels: Discussed seperately, it is correct.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageProcessingResult.cs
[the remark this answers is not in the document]
stijnmoreels: Exceptions locations will be handled separately in a diff session.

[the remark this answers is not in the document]
stijnmoreels: Will be discussed later in #456

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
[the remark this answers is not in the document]
stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/453/commits/337000c9c15c1589c905351e24cb823804128815

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/453/commits/0f1aed451b4444aa8e13f18aba73171cc6558007

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
```diff
@@ -23,12 +23,78 @@ public interface IMessagePumpCircuitBreaker
...
+        public bool IsClosed => _state is CircuitBreakerState.Closed;
+
+        /// <summary>
+        /// Gets the boolean flag that indicates whether the message pump is in the 'Open' circuit breaker state,
+        /// and should stop retrieving messages.
+        /// </summary>
+        public bool IsOpen => _state is CircuitBreakerState.Open;
+
+        /// <summary>
+        /// Gets the boolean flag tht indicates whether the message pump is in the 'Half-Open' circuit breaker state,
```
fgheysels, line 67:
```suggestion
        /// Gets the boolean flag that indicates whether the message pump is in the 'Half-Open' circuit breaker state,
```
typo :)

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
```diff
@@ -23,12 +23,78 @@ public interface IMessagePumpCircuitBreaker
...
+        /// and should try retrieving a single message at a time before fully starting retrieving messages.
+        /// </summary>
```
fgheysels, line 69: Minor remark, but if the pump is in a certain state, it is already working as defined by that state.
The comments on these properties say that 'the pump should ....'  , but 'the pump is ' actually already behaving in that way.
So, I'd change the wording of the comments in the XML comments to something like this :)

> Gets a boolean that indicates whether the message pump is in a 'Half-Open' state or not.

and then maybe in the remarks section some explanation:

> When the message pump is in the 'Half-Open' state, the pump retrieves a single message and verifies if the message can be processed.  If the message is processed succesfully, the pump will transition into the 'Closed' state.

fgheysels: (And then some similar comments for the `IsClosed` and `IsOpen` properties :)

stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/453/commits/bba3b25371bf6f05f45dd9179c57a2534d5c74c9

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
```diff
@@ -23,12 +23,86 @@ public interface IMessagePumpCircuitBreaker
...
+        {
+            _state = state;
+
+            Options = options ?? throw new ArgumentNullException(nameof(options));
+        }
 
         /// <summary>
-        /// Continue the process of receiving messages in the message pump after a successful message handling.
+        /// Gets the boolean flag that indicates whether the message pump is in the 'Closed' circuit breaker state or not.
+        /// and should start retrieving messages.
```
fgheysels, line 56: L56 can be removed :)

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
```diff
@@ -23,12 +23,86 @@ public interface IMessagePumpCircuitBreaker
...
-        /// <param name="jobId">The unique identifier to distinguish the message pump in the application services.</param>
-        /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
-        Task ResumeMessageProcessingAsync(string jobId);
+        /// <remarks>
+        ///     When the message pump is in the 'Closed' state, the pump is retrieving messages at a normal rate.
+        /// </remarks>
+        public bool IsClosed => _state is CircuitBreakerState.Closed;
+
+        /// <summary>
+        /// Gets the boolean flag that indicates whether the message pump is in the 'Open' circuit breaker state or not.
```
fgheysels, line 64:
```suggestion
        /// Gets a boolean flag that indicates whether the message pump is in the 'Open' circuit breaker state or not.
```

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
```diff
@@ -23,12 +23,86 @@ public interface IMessagePumpCircuitBreaker
...
+        private MessagePumpCircuitState(CircuitBreakerState state, MessagePumpCircuitBreakerOptions options)
+        {
+            _state = state;
+
+            Options = options ?? throw new ArgumentNullException(nameof(options));
+        }
 
         /// <summary>
-        /// Continue the process of receiving messages in the message pump after a successful message handling.
+        /// Gets the boolean flag that indicates whether the message pump is in the 'Closed' circuit breaker state or not.
```
fgheysels, line 55:
```suggestion
        /// Gets a boolean flag that indicates whether the message pump is in the 'Closed' circuit breaker state or not.
```

src/Arcus.Messaging.Pumps.Abstractions/Resiliency/IMessagePumpCircuitBreaker.cs
stijnmoreels:
```suggestion
```

PR #454, opened 2024-11-26
Spike: Fix circuitbreaker functionality

fgheysels:


fgheysels:
Discussed this offline and we decided to continue with the approach that is proposed in this PR:
#453

PR #461, opened 2024-12-27
feat: add circuit breaker event handlers

### LINKED ISSUE #459: Provide extension-points on the MessagePump or CircuitBreaker where we can 'hook in' when the CircuitBreaker's state changes
The ServiceBus CircuitBreaker can change from Closed to Open to Half-Open to Closed.
It can be meaningful if an application developer can be notified when the state of the CircuitBreaker changes.

See also this discussion in the CircuitBreaker PR:
https://github.com/arcus-azure/arcus.messaging/pull/453#discussion_r1872834729

stijnmoreels:
Provide an extension point via event handlers so that developers can get notified on any state transition changes of their registered message pump.

Closes #459

### DIFF
--- a/docs/preview/02-Features/06-general-messaging.md
+++ b/docs/preview/02-Features/06-general-messaging.md
@@ -78,6 +78,31 @@ Both the recovery period after the circuit is open and the interval between mess
 }
 ```
 
+#### 🔔 Get notified on a circuit breaker state transition
+Transitions from circuit breaker states happens internally and automatically. The library supports a notification system that lets you register event handlers that gets called upon a state transition.
+
+After the message pump and/or message handlers are registered, you can add one or more handlers linked to specific transitions to the previously registered pump.
+
+```csharp
+using Arcus.Messaging.Pumps.Abstractions;
+
+services.AddServiceBusMessagePump(...)
+        .WithCircuitBreakerEventHandler<MyFirstCircuitBreakerEventHandler>()
+        .WithCircuitBreakerEventHandler<MySecondCircuitBreakerEventHandler>();
+```
+
+The instances should implement the `ICircuitBreakerEventHandler`, which allows you to access the new state for the pump.
+
+```csharp
+public class MyFirstCircuitBreakerEventHandler : ICircuitBreakerEventHandler
+{
+    public Task OnTransitionAsync(MessagePumpCircuitState newState)
+    {
+        // ...
+    }
+}
+```
+
 ### Pause message processing for a fixed period of time 
 ⚡ If you use one of the message-type specific packages like `Arcus.Messaging.Pumps.EventHubs`, you will automatically get this functionality. If you implement your own message pump, please use the `services.AddMessagePump(...)` extension which makes sure that you also registers this functionality.
 
--- a/docs/preview/02-Features/06-general-messaging.md
+++ b/docs/preview/02-Features/06-general-messaging.md
@@ -87,8 +87,8 @@ After the message pump and/or message handlers are registered, you can add one o
 using Arcus.Messaging.Pumps.Abstractions;
 
 services.AddServiceBusMessagePump(...)
-        .WithCircuitBreakerEventHandler<MyFirstCircuitBreakerEventHandler>()
-        .WithCircuitBreakerEventHandler<MySecondCircuitBreakerEventHandler>();
+        .WithCircuitBreakerStateChangedEventHandler<MyFirstCircuitBreakerEventHandler>()
+        .WithCircuitBreakerStateChangedEventHandler<MySecondCircuitBreakerEventHandler>();
 ```
 
 The instances should implement the `ICircuitBreakerEventHandler`, which allows you to access the new state for the pump.
--- a/src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
+++ b/src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
@@ -53,17 +53,17 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <summary>
         /// Adds an <see cref="ICircuitBreakerEventHandler"/> implementation for a specific message pump to the application services.
         /// </summary>
-        /// <typeparam name="THandler">The custom type of the event handler.</typeparam>
+        /// <typeparam name="TEventHandler">The custom type of the event handler.</typeparam>
         /// <param name="services">The application services to register the event handler.</param>
         /// <param name="jobId">The unique ID to distinguish the message pump to register this event handler for.</param>
         /// <param name="implementationFactory">The factory function to create the custom <see cref="ICircuitBreakerEventHandler"/> implementation.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="services"/> or <paramref name="implementationFactory"/> is <c>null</c>.</exception>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
-        public static IServiceCollection AddCircuitBreakerEventHandler<THandler>(
+        public static IServiceCollection AddCircuitBreakerEventHandler<TEventHandler>(
             this IServiceCollection services,
             string jobId,
-            Func<IServiceProvider, THandler> implementationFactory)
-            where THandler : ICircuitBreakerEventHandler
+            Func<IServiceProvider, TEventHandler> implementationFactory)
+            where TEventHandler : ICircuitBreakerEventHandler
         {
             if (services is null)
             {
--- a/src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
+++ b/src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
@@ -32,22 +32,22 @@ namespace Microsoft.Extensions.DependencyInjection
             services.TryAddSingleton<IMessagePumpLifetime, DefaultMessagePumpLifetime>();
             services.TryAddSingleton<IMessagePumpCircuitBreaker>(
                 provider => new DefaultMessagePumpCircuitBreaker(provider, provider.GetService<ILogger<DefaultMessagePumpCircuitBreaker>>()));
-            
+
             return services.AddHostedService(implementationFactory);
         }
 
         /// <summary>
         /// Adds an <see cref="ICircuitBreakerEventHandler"/> implementation for a specific message pump to the application services.
         /// </summary>
-        /// <typeparam name="THandler">The custom type of the event handler.</typeparam>
+        /// <typeparam name="TEventHandler">The custom type of the event handler.</typeparam>
         /// <param name="services">The application services to register the event handler.</param>
         /// <param name="jobId">The unique ID to distinguish the message pump to register this event handler for.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="services"/> is <c>null</c>.</exception>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
-        public static IServiceCollection AddCircuitBreakerEventHandler<THandler>(this IServiceCollection services, string jobId)
-            where THandler : ICircuitBreakerEventHandler
+        public static IServiceCollection AddCircuitBreakerEventHandler<TEventHandler>(this IServiceCollection services, string jobId)
+            where TEventHandler : ICircuitBreakerEventHandler
         {
-            return AddCircuitBreakerEventHandler(services, jobId, provider => ActivatorUtilities.CreateInstance<THandler>(provider));
+            return AddCircuitBreakerEventHandler(services, jobId, provider => ActivatorUtilities.CreateInstance<TEventHandler>(provider));
         }
 
         /// <summary>
--- a/src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
+++ b/src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
@@ -35,5 +35,52 @@ namespace Microsoft.Extensions.DependencyInjection
             
             return services.AddHostedService(implementationFactory);
         }
+
+        /// <summary>
+        /// Adds an <see cref="ICircuitBreakerEventHandler"/> implementation for a specific message pump to the application services.
+        /// </summary>
+        /// <typeparam name="THandler">The custom type of the event handler.</typeparam>
+        /// <param name="services">The application services to register the event handler.</param>
+        /// <param name="jobId">The unique ID to distinguish the message pump to register this event handler for.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="services"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
+        public static IServiceCollection AddCircuitBreakerEventHandler<THandler>(this IServiceCollection services, string jobId)
+            where THandler : ICircuitBreakerEventHandler
+        {
+            return AddCircuitBreakerEventHandler(services, jobId, provider => ActivatorUtilities.CreateInstance<THandler>(provider));
+        }
+
+        /// <summary>
+        /// Adds an <see cref="ICircuitBreakerEventHandler"/> implementation for a specific message pump to the application services.
+        /// </summary>
+        /// <typeparam name="THandler">The custom type of the event handler.</typeparam>
+        /// <param name="services">The application services to register the event handler.</param>
+        /// <param name="jobId">The unique ID to distinguish the message pump to register this event handler for.</param>
+        /// <param name="implementationFactory">The factory function to create the custom <see cref="ICircuitBreakerEventHandler"/> implementation.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="services"/> or <paramref name="implementationFactory"/> is <c>null</c>.</exception>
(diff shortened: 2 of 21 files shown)


src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -146,28 +156,55 @@ protected async Task WaitMessageIntervalDuringRecoveryAsync(CancellationToken ca
...
             CircuitState = MessagePumpCircuitState.Closed;
+
+            await NotifyCircuitBreakerEventHandlersAsync();
+        }
+
+        private async Task NotifyCircuitBreakerEventHandlersAsync()
+        {
+            ICircuitBreakerEventHandler[] eventHandlers = GetEventHandlersForPump();
+
+            await Task.WhenAll(eventHandlers.Select(h => h.OnTransitionAsync(CircuitState)));
```
stijnmoreels, line 199: Was wondering if this is a good place to use `Task.WhenAny` instead 🤔

src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -35,5 +35,52 @@ public static IServiceCollection AddMessagePump<TMessagePump>(
...
+            {
+                throw new ArgumentNullException(nameof(implementationFactory));
+            }
+
+            if (string.IsNullOrWhiteSpace(jobId))
+            {
+                throw new ArgumentException("Requires a non-blank job ID to distinguish the message pump for which this circuit breaker event handler is registered", nameof(jobId));
+            }
+
+            return services.AddTransient(provider => new CircuitBreakerEventHandler(jobId, implementationFactory(provider)));
```
stijnmoreels, line 83: Made this transient, just so that even if people incorrectly use states, we get a fresh handler.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -146,28 +156,55 @@ protected async Task WaitMessageIntervalDuringRecoveryAsync(CancellationToken ca
             Logger.LogDebug("Circuit breaker caused message pump '{JobId}' to wait message interval during recovery of '{Interval}' during the '{State}' state", JobId, CircuitState.Options.MessageIntervalDuringRecovery.ToString("g"), CircuitState);
             await Task.Delay(CircuitState.Options.MessageIntervalDuringRecovery, cancellationToken);
 
+            MessagePumpCircuitState oldState = CircuitState;
+            bool shouldTriggerTransitionEvent = !oldState.IsHalfOpen;
+
             CircuitState = CircuitState.TransitionTo(CircuitBreakerState.HalfOpen);
+
+            if (shouldTriggerTransitionEvent)
```
stijnmoreels, line 164: Don't notify when we go from half-open to half-open.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
fgheysels: There's a difference between the 2; `WhenAll` waits 'till all tasks are completed, `WhenAny` stops waiting as soon as one task has completed.
I wonder why `WaitAny` would be appropriate ?

fgheysels: In hindsight ....  indeed, is it necessary to await on those eventhandlers ?  Is it even necessary to await on any of them ?  Isn't this just 'fire and forget' cfr events in Windows applications ?

docs/preview/02-Features/06-general-messaging.md
```diff
@@ -78,6 +78,31 @@ Both the recovery period after the circuit is open and the interval between mess
...
+#### 🔔 Get notified on a circuit breaker state transition
+Transitions from circuit breaker states happens internally and automatically. The library supports a notification system that lets you register event handlers that gets called upon a state transition.
+
+After the message pump and/or message handlers are registered, you can add one or more handlers linked to specific transitions to the previously registered pump.
+
+```csharp
+using Arcus.Messaging.Pumps.Abstractions;
+
+services.AddServiceBusMessagePump(...)
+        .WithCircuitBreakerEventHandler<MyFirstCircuitBreakerEventHandler>()
```
fgheysels, line 90: Should we specify what kind of event-handler this is ?  Or make it clear that this is an event-handler that is triggered when the CircuitBreakerState has changed ?

`WithCircuitBreakerStateChangedEventHandler` or something ?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
fgheysels: Actually, is it even necessary to call `TransitionTo` if we remain in half-open state ?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Only not sure what happens with remaining tasks in `Task.WhenAny`. Thought that was mostly used for cases where you have more than one way of doing things and just wait for any of them to complete, but not all complete in the end.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Ah, that's a good point. Now that we have the `IsHalfOpen`, we can check for that. Thx! 👍

docs/preview/02-Features/06-general-messaging.md
stijnmoreels: Could be, but than the interface `ICircuitBreakerEventHandler` should also change, I believe.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
fgheysels: But, do we have to wait on any of those tasks ?   Is the circuitbreaker interested in what happens by those tasks ?
Or can we treat them as some kind of eventhandler ?   The circuitbreaker notifies that something happened (change of state), and anyone who's interested can act upon it.  But is the circuitbreaker interested in the outcome of that ?

docs/preview/02-Features/06-general-messaging.md
fgheysels: I'm not sure of that.  Maybe we can have multiple 'events' in the future (when the state changes, when something else changes), but these type of events can all be handled by `ICircuitBreakerEventHandler` types ?  The only thing that's different is when they're triggered (the event that causes them to trigger)

stijnmoreels: Yes, but the interface has a `TransitionTo` method, right? Isn't that always something that will be needed for when 'state changes'?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Done that here: https://github.com/arcus-azure/arcus.messaging/pull/461/commits/6a24ad8e45a911c47b3652f53f0302e2e104a1b5

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
fgheysels: I think we should not wait on any of those eventhandlers.
I'm even wondering / doubting if we should define the `OnTransitionAsync` method as a method that returns a Task.

docs/preview/02-Features/06-general-messaging.md
fgheysels: When State changes, we'll probably indeed use `TransitionTo`.
However, I wouldn't tightly couple the name of the interface to the name of the property and keep the name of the interface `ICircuitBreakerEventHandler`.

stijnmoreels: Done in https://github.com/arcus-azure/arcus.messaging/pull/461/commits/3ebfe3c720d95b7a1412d13a7fd2e28fb76dc9f4

src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -35,5 +35,52 @@ public static IServiceCollection AddMessagePump<TMessagePump>(
...
+        /// <param name="services">The application services to register the event handler.</param>
+        /// <param name="jobId">The unique ID to distinguish the message pump to register this event handler for.</param>
+        /// <param name="implementationFactory">The factory function to create the custom <see cref="ICircuitBreakerEventHandler"/> implementation.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="services"/> or <paramref name="implementationFactory"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
+        public static IServiceCollection AddCircuitBreakerEventHandler<TEventHandler>(
+            this IServiceCollection services,
+            string jobId,
+            Func<IServiceProvider, TEventHandler> implementationFactory)
+            where TEventHandler : ICircuitBreakerEventHandler
```
fgheysels, line 66: Just wondering ....

Why `AddCircuitBreakerEventHandler<TEventHandler> .... where TEventHandler : ICircuitBreakerEventHandler`  instead of `AddCircuitBreakerEventHandler<ICircuitBreakerEVentHandler>()` ?

src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -35,5 +35,52 @@ public static IServiceCollection AddMessagePump<TMessagePump>(
...
+        /// <summary>
+        /// Adds an <see cref="ICircuitBreakerEventHandler"/> implementation for a specific message pump to the application services.
+        /// </summary>
+        /// <typeparam name="THandler">The custom type of the event handler.</typeparam>
+        /// <param name="services">The application services to register the event handler.</param>
+        /// <param name="jobId">The unique ID to distinguish the message pump to register this event handler for.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="services"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
+        public static IServiceCollection AddCircuitBreakerEventHandler<THandler>(this IServiceCollection services, string jobId)
+            where THandler : ICircuitBreakerEventHandler
```
fgheysels, line 48: Is there a reason why the type-parameter here is called `THandler` whereas in the method below it is named `TEventHandler` ?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -168,6 +182,26 @@ protected void NotifyResumeRetrievingMessages()
             Logger.LogDebug("Circuit breaker caused message pump '{JobId}' to transition back from '{CurrentState}' to a 'Closed' state, retrieving messages is resumed", JobId, CircuitState);
 
             CircuitState = MessagePumpCircuitState.Closed;
+
+            NotifyCircuitBreakerEventHandlers();
+        }
+
+        private void NotifyCircuitBreakerEventHandlers()
```
fgheysels, line 189: I'm nitpicking, but we actually **trigger** the eventhandlers and by doing so, we **notify** the subscribers. 
So, I think this method should be named `TriggerCircuitBreakerStateChangedEventHandlers` or `NotifyCircuitBreakerStateChangedSubscribers`

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
```diff
@@ -168,6 +182,26 @@ protected void NotifyResumeRetrievingMessages()
...
+        {
+            ICircuitBreakerEventHandler[] eventHandlers = GetEventHandlersForPump();
+
+            foreach (var handler in eventHandlers)
+            {
+                Task.Run(() => handler.OnTransition(CircuitState));
```
fgheysels, line 195: I was wondering if this is a good way to trigger the eventhandlers in a fire-and-forget manner, and I came accross this interesting article:
https://michaelceber.medium.com/net-core-web-api-how-to-correctly-fire-and-forget-database-commands-from-controller-end-point-cad7a879630

The solution that is proposed there, offers the ability for the event-subscribers to access services that are registered in the DI container in a safe way.  Might be something to consider ?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: That was done here: https://github.com/arcus-azure/arcus.messaging/pull/461/commits/f43b305233c78d62bb9d754b9e4e23f35c595acf

src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: You mean without the extra generic condition? Why would you remove that?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: The blog talks mostly about moving the `Task.Run(...)` to an seperate service to not mix/match it with app code -> agree with that; but functionally, it is the same, though? Not sure why that would be an improvement as we're not showing any of this code to the users.

src/Arcus.Messaging.Pumps.Abstractions/Extensions/IServiceCollectionExtensions.cs
fgheysels: Nevermind ....    Seems like I've been away from writing code for too long.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: That can also be done now, right? There is an overload for the event handler that lets you either regiser the type or provide an implementation factory function with the service-provider.

src/Arcus.Messaging.Tests.Integration/Fixture/WorkerOptions.cs
```diff
@@ -78,30 +79,33 @@ public WorkerOptions ConfigureSerilog(Action<LoggerConfiguration> configure)
...
-                           {
-                               config.WriteTo.XunitTestLogging(_outputWriter);
-                           }
-
-                           foreach (Action<LoggerConfiguration> configure in _additionalSerilogConfigOptions)
-                           {
-                               configure(config);
-                           }
-                       });
+                       .UseSerilog(Log.Logger);
```
stijnmoreels, line 108: After some investigation and step-by-step pinpointing, I have reduced the space of the problem to Serilog and its registration. I changed it to this, more explicit, way which finally made the test work again.

fgheysels: Weird.  Do you have any idea what the exact reason was why it failed ?

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
fgheysels: And how can the service-provider be used within the eventhandler ?  That's not entirely clear to me atm.

stijnmoreels: The event handlers are registered via the `IServiceCollection`. So, `.WithCircuitBreakerStateChangedEventHandler<T>(…)` has two overloads:
* One without additional arguments;
* One with an implementation factory function: `Func<IServiceProvider, TEventHandler>`.

When you register such an event handler, you have direct access to the service-provider.

```csharp
services.AddServiceBusMessagePump(…)
        .WithCircuitBreakerStateChangedEventHandler((IServiceProvider provider) =>
        {
            return new MyCircuitBreakerEventHandler(provider, …);
        });
```

src/Arcus.Messaging.Tests.Integration/Fixture/WorkerOptions.cs
stijnmoreels: The test has two Serilog sinks; 1 to stream logs to the test output, 1 to save to an in-memory App Insights sink to later assert on the telemetry.
The test in particular sometimes failed in the past (only remotely), but recently, it went smoothly again, until now where it failed constantly. The exact reason, I don’t know. It has to be environment-specific, I’m thinking about VM’s and versions, even. The test never failed to run locally, even together with other tests.

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: Done that here: https://github.com/arcus-azure/arcus.messaging/pull/461/commits/5fef61ad7cefd3d09ed2c02cdf63cca4982313e5

src/Arcus.Messaging.Pumps.Abstractions/MessagePump.cs
stijnmoreels: I would think that this is common knowledge as we follow the same practice as with Microsoft-developed services, which we follow for all our message handlers as well. Think this is even .NET Core library development best practices.

I would think that it would be too specific for feature docs? Especially since most of the time you don't need to access the service provider directly and can just inject?

src/Arcus.Messaging.Tests.Integration/Fixture/WorkerOptions.cs
fgheysels: Weird, but good that it's working again :)

fgheysels (approved):
Thanks for this !
I'll be happy to use it on our customer project!

PR #463, opened 2024-12-27
chore: slim down used nuget packages

stijnmoreels:
Slim down the used NuGet packages throughout the solution, both transient and unused.


fgheysels (approved):

fgheysels:
Can we remove those without having to make changes in the code ?  This means that those references were actually unnecessary ?

stijnmoreels:
> Can we remove those without having to make changes in the code ? This means that those references were actually unnecessary ?

Yes, some where added in the past due to package version contrains, other's were added to update transient dependencies, but only some of them are necessary.

fgheysels:
Can this PR be published, or does it require additional work ?

stijnmoreels:
> Can this PR be published, or does it require additional work ?

Will need to merge with `main` to get the fix, but otherwise, I think it's almost done.

Issue #449, opened 2024-08-23
Phase out the use of Guard.NET


fgheysels:
We'll remove the dependency Arcus.Messaging has on the Guard.NET package.
See [this](https://github.com/arcus-azure/arcus/discussions/362) discussion.

stijnmoreels:
All referencees to Guard.NET has been removed, only transient dependencies remain.

PR #466, opened 2025-01-29
chore: remove guard.net references from eventhubs

stijnmoreels:
Remove all calls that use the Guard.NET library in the EventHubs-related projects.

Relates to #449


fgheysels (approved):
No changes in the csproj files yet to remove the package ?

stijnmoreels:
> No changes in the csproj files yet to remove the package ?

Are transient via Security and Observability, so no, not yet.

PR #467, opened 2025-01-30
chore: remove guard.net references from servicebus

stijnmoreels:
Remove any calls to the Guard.NET library in ServiceBus-related projects.

Relates to #449

### DIFF
--- a/src/Arcus.Messaging.Pumps.ServiceBus/Arcus.Messaging.Pumps.ServiceBus.csproj
+++ b/src/Arcus.Messaging.Pumps.ServiceBus/Arcus.Messaging.Pumps.ServiceBus.csproj
@@ -29,7 +29,6 @@
   <ItemGroup>
     <PackageReference Include="Arcus.Security.Providers.AzureKeyVault" Version="[2.0.0,3.0.0)" />
     <PackageReference Include="Azure.Identity" Version="1.12.0" />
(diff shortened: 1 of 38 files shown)


src/Arcus.Messaging.Pumps.ServiceBus/Arcus.Messaging.Pumps.ServiceBus.csproj
```diff
@@ -29,7 +29,6 @@
   <ItemGroup>
     <PackageReference Include="Arcus.Security.Providers.AzureKeyVault" Version="[2.0.0,3.0.0)" />
     <PackageReference Include="Azure.Identity" Version="1.12.0" />
```
fgheysels, line 31: I think we should upgrade Azure.Identity to a more recent version.
There are quite often new versions released of this package, so I was wondering if we should define this dependency as `Version=1.*` ?

stijnmoreels: I would opt for version ranges, as it shows limitations at compile time io at runtime.

stijnmoreels: But this is a bit an out-of-scope discussions as it spans all used Microsoft dependencies.

PR #471, opened 2025-02-11
chore(deps): deprecate arcus.security references in service-bus projects

stijnmoreels:
* Deprecate all references to `Arcus.Security` types in the Service bus-related projects with a deprecation message on what to do instead.
* Remove all mentioning of the Arcus secret store in the Service bus-related documentation, and instead promote managed identity.

This is part of the excerise described in #470 where the Messaging repo becomes more lightweight and less dependent on other packages.


fgheysels:
I think we should not only support ManagedIdentity to connect to ServiceBus.
While we can remove the dependency to Arcus.Security, I think we should still have a way to connect to ServiceBus by specifying a clientid / clientsecret for instance.  But, instead of passing in a secretname and use that secretname to retrieve the secret from KeyVault using Arcus.Security, we could pass in the secret directly.
It's then the responsibility of the client application to retrieve the secret from a secure location and use it.

stijnmoreels:
> I think we should not only support ManagedIdentity to connect to ServiceBus. While we can remove the dependency to Arcus.Security, I think we should still have a way to connect to ServiceBus by specifying a clientid / clientsecret for instance. But, instead of passing in a secretname and use that secretname to retrieve the secret from KeyVault using Arcus.Security, we could pass in the secret directly. It's then the responsibility of the client application to retrieve the secret from a secure location and use it.

That's still kinda possible as we still support the `IConfiguration` overload, where people can register their own source/authentication and retrieve it there.

fgheysels:
> > I think we should not only support ManagedIdentity to connect to ServiceBus. While we can remove the dependency to Arcus.Security, I think we should still have a way to connect to ServiceBus by specifying a clientid / clientsecret for instance. But, instead of passing in a secretname and use that secretname to retrieve the secret from KeyVault using Arcus.Security, we could pass in the secret directly. It's then the responsibility of the client application to retrieve the secret from a secure location and use it.
> 
> That's still kinda possible as we still support the `IConfiguration` overload, where people can register their own source/authentication and retrieve it there.

I'd have to take a look at what you mean exactly.
My idea would just be to provide a method that allows you to specify the clientid / clientsecret itself.   Arcus should not be responsible for retrieving that information.  The program / component that hosts Arcus.Messaging should be responsible.

fgheysels:
> > I'd have to take a look at what you mean exactly.
> > My idea would just be to provide a method that allows you to specify the clientid / clientsecret itself. Arcus should not be responsible for retrieving that information. The program / component that hosts Arcus.Messaging should be responsible.
> 
> Currently supported, even after this PR:
> 
> ```cs
> Host.CreateDefaultBuilder()
>     .ConfigureAppConfiguration(config => config.AddAzureKeyVault(new Uri(...), new ClientSecretCredential(...))
>     .ConfigureServices(services =>
>     {
>         services.AddServiceBusTopicMessagePump("<subscription>", (IConfiguration config) =>
>         {
>             return config["Topic:ConnectionString:From:IConfig:Via:KeyVault");
>         });
>     });
> ```

Yeah, but it's a pity you have to use it like that if your program that is using Arcus.Messaging is also using Arcus.Security.
In such case, I think it would be better if we can directly pass in the clientid/clientsecret.
I'l have to peek into the code to see if I can provide an example of what I mean.

stijnmoreels:
> Yeah, but it's a pity you have to use it like that if your program that is using Arcus.Messaging is also using Arcus.Security. In such case, I think it would be better if we can directly pass in the clientid/clientsecret. I'l have to peek into the code to see if I can provide an example of what I mean.

Think that is a bit the purpose, that people see that they need to migra5to something else.
Happy to discuss this separately.

fgheysels:
> > I'd have to take a look at what you mean exactly.
> > My idea would just be to provide a method that allows you to specify the clientid / clientsecret itself. Arcus should not be responsible for retrieving that information. The program / component that hosts Arcus.Messaging should be responsible.
> 
> Currently supported, even after this PR:
> 
> ```cs
> Host.CreateDefaultBuilder()
>     .ConfigureAppConfiguration(config => config.AddAzureKeyVault(new Uri(...), new ClientSecretCredential(...))
>     .ConfigureServices(services =>
>     {
>         services.AddServiceBusTopicMessagePump("<subscription>", (IConfiguration config) =>
>         {
>             return config["Topic:ConnectionString:From:IConfig:Via:KeyVault");
>         });
>     });
> ```

True.
But, then you're forced to store a connectionstring (secret) in your configuration.  One of the reasons we have Arcus.Security is to make a distinction between secrets and configuration.
When you want to go that route, you'll need to add your secret-store (KeyVault for instance) as a Configuration-source (which is possible, but do we want that ?).  Another solution is to use KeyVault-references in your Configuration (which is also a valid way of working but then again, no explicit distinction between secrets & configuration).

I do understand that you want to remove the dependency to Arcus.Security though. 
Therefore, as an alternative for teams that use Arcus.Messaging and Arcus.Security, I would provide an option to pass in the connection-string directly.
It's then the responsability of the program that uses those libraries to retrieve the connectionstring from the secret-store and pass it to Arcus.Messaging.  Something like this:

```csharp
string connectionstring = await secretProvider.GetRawSecretAsync(mySecretName);

services.AddServiceBusQueueMessagePump(connectionString);
```

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
```diff
@@ -24,6 +24,51 @@ public class AzureServiceBusMessagePumpSettings
...
+        /// <param name="options">The options that influence the behavior of the <see cref="AzureServiceBusMessagePump"/>.</param>
+        /// <param name="serviceProvider">The collection of services to use during the lifetime of the <see cref="AzureServiceBusMessagePump"/>.</param>
+        /// <exception cref="ArgumentNullException">
+        /// Thrown when the <paramref name="getConnectionStringFromConfigurationFunc"/>, <paramref name="options"/> or <paramref name="serviceProvider"/> is <c>null</c>.
+        /// </exception>
+        public AzureServiceBusMessagePumpSettings(
+            string entityName,
+            string subscriptionName,
+            ServiceBusEntityType serviceBusEntity,
+            Func<IConfiguration, string> getConnectionStringFromConfigurationFunc,
```
fgheysels, line 43: I'm not a fan of having to retrieve the connctionstring from config.
Why not just pass in the connectionstring as a string ?  Its the responsibility of the user to retrieve it from any location he wants.

src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
```diff
@@ -29,6 +29,7 @@ public static class AzureClientFactoryBuilderExtensions
         /// <exception cref="ArgumentException">Thrown when the <paramref name="connectionStringSecretName"/> is blank.</exception>
         /// <exception cref="InvalidOperationException">Thrown when the Arcus secret store is not registered.</exception>
         /// <exception cref="SecretNotFoundException">Thrown when no Azure EventHubs connection string secret was found in the Arcus secret store.</exception>
+        [Obsolete("Will be removed in v3.0, please use Microsoft's built-in Azure SDK clients to register a " + nameof(ServiceBusClient) + " to remove the " + nameof(ISecretProvider) + " requirement")]
```
fgheysels, line 32: I believe we can make the entire class obsolete ?

stijnmoreels:
Best to discus this separately: 'What to do with projects that use Arcus.Security in combination with Arcus.Messaging.'

src/Arcus.Messaging.ServiceBus.Core/Extensions/AzureClientFactoryBuilderExtensions.cs
stijnmoreels: Yes, probably in time, but the scope of this PR is only to deprecate Arcus.Security references.

stijnmoreels: Oh, my mistake, thought that this was another extensions class, yes we can.

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
fgheysels: In other words, my idea would be to introduce a method that adds a messagepump and that accepts the connectionstring that must be used as a parameter:

```
public static ServiceBusMessageHandlerCollection AddServiceBusQueueMessagePump (
      this IServiceCollection services,
      string connectionString,
      string queueName,
      Action<IAzureServiceBusQueueMessagePumpOptions> options) {}
```
Maybe also provide a version that doesn't have the queueName parameter, because it's possible that the queue to which to connect is provided in the connectionstring as well.
(And for topics as well).

fgheysels:
> Best to discus this separately: 'What to do with projects that use Arcus.Security in combination with Arcus.Messaging.’
> 
> 3. Figuring out what to expose as functionality is critical, as this has to be modular enough. (The thing you proposed of having the connection string directly, would mean yet an extra extension for an already too-big extension set for registering message pumps. Also: retrieving a secret store at that point isn’t really possible in a safe way, I believe, as the pump is registered at the same time as the secret store: through the services).

Actually, from a security perspective it's good practice to minimize the need / usage of secrets.  Then, we might consider to only support establishing a connection with managed identity.

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
stijnmoreels: Maybe an idea to have a `Func<IServiceProvider, ServiceBusClient>`? That would also make it possible to still use the secret store.

fgheysels: Ok, I agree.
Just wondering why you would want to return a `ServiceBusClient` instead of a string (connectionstring).  Don't we expose the internal workings then ? And make it harder too adapt to change afterwards ?

stijnmoreels: I would even opt to later move to use the `ServiceBusClient`s registered by the Azure SDK extensions (like `services.AddAzureClients(clients => clients.AddServiceBusClient(…))`) instead of us creating and managing these clients ourselves. So, having this delegate would already provide people with using those built-in clients already.
Exposing inner workings, that’s true. Although in this case it is not an inner working of our own creation (like if we would expose how custom message handlers are registered, or how message routing works), but a Microsoft known type (like if we would expose `TokenCredential`).
But, still true, it exposes a bit more than before.

stijnmoreels: I've created something locally as a proposal, we can go over it separately to see if it fits with you had in mind.

fgheysels: Yes, I'll schedule a meeting for it.

docs/preview/02-Features/06-general-messaging.md
```diff
@@ -86,7 +86,7 @@ After the message pump and/or message handlers are registered, you can add one o
 ```csharp
 using Arcus.Messaging.Pumps.Abstractions;
 
-services.AddServiceBusMessagePump(...)
+services.AddServiceBusMessagePumpUsingManagedIdentity(...)
```
stijnmoreels, line 89:
```suggestion
services.AddServiceBusMessagePump(...)
```

docs/preview/02-Features/06-general-messaging.md
```diff
@@ -127,10 +127,10 @@ public class Program
 {
     public void ConfigureServices(IServiceCollection services)
     {
-        services.AddServiceBusMessagePump(..., options => options.JobId = "abc-123")
+        services.AddServiceBusMessagePumpUsingManagedIdentity(..., options => options.JobId = "abc-123")
```
stijnmoreels, line 130:
```suggestion
        services.AddServiceBusMessagePump(..., options => options.JobId = "abc-123")
```

docs/preview/02-Features/06-general-messaging.md
```diff
@@ -127,10 +127,10 @@ public class Program
...
                 .WithServiceBusMessageHandler<..., ...>();
 
-        services.AddEventHubsMessagePump(..., options => options.JobId = "def-456")
+        services.AddEventHubsMessagePumpUsingManagedIdentity(..., options => options.JobId = "def-456")
```
stijnmoreels, line 133:
```suggestion
        services.AddEventHubsMessagePump(..., options => options.JobId = "def-456")
```

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpSettings.cs
fgheysels: Discussed this offline.
2 follow PR's will be created:
- one where we introduce the functionality to establish a connection by offering a factory that gives access to the ServiceProvider and is responsible for returning a `ServiceBusClient` instance.  Next to that, an extension method will be created that accepts a `TokenCredential` which is then used by Arcus.Messaging to establish the connection to ServiceBus
- another PR where we deprecate the functionality that allows the MessagePump to create its own servicebus-topic-subscription

PR #472, opened 2025-02-11
chore(deps): use system.text.json instead of newtonsoft.json throughout

stijnmoreels:
Use Microsoft's built-in `System.Text.Json` instead of adding `Newtonsoft.Json` reference in all Messaging-projects.

This is part of the lightweight exercise described in #470 where dependencies are being investigated on their added-value and maintainence cost.


src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
```diff
@@ -29,7 +29,10 @@
   <ItemGroup>
     <PackageReference Include="Arcus.Observability.Telemetry.Serilog.Enrichers" Version="[3.0.0,4.0.0)" />
     <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.21.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+
+    <!-- TODO: will be removed in v3.0 -->
+    <PackageReference Include="System.Text.Json" Version="[8.0.5,10.0.0)" Condition="'$(TargetFramework)' != 'net8.0'" />
+    <!-- TODO: will be removed in v3.0 -->
```
fgheysels, line 35: What do you mean ?  Why will we remove this dependency in a later version ?

stijnmoreels: As it is only used by .NET Standard and.NET 6. And we are migrating away from them, I believe.

stijnmoreels: It's included by default 8n .NET 8, but not in the other frameworks.

fgheysels: But, should we then have a version range 'til 10.x ?

src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouter.cs
```diff
@@ -477,66 +476,40 @@ protected virtual bool TryDeserializeToMessageFormat(string message, Type messag
...
-            JsonSerializer jsonSerializer = CreateJsonSerializer();
-            EventHandler<ErrorEventArgs> eventHandler = (sender, args) =>
-            {
-                success = false;
-                Logger.LogTrace(args.ErrorContext.Error, "Incoming message failed to be JSON deserialized to message type '{MessageType}' at {Path}", messageType.Name, args.ErrorContext.Path);
-                args.ErrorContext.Handled = true;
-            };
-            jsonSerializer.Error += eventHandler;
-            
+            JsonSerializerOptions options = CreateJsonSerializerOptions();
```
fgheysels, line 479: Do we want to create a new instance of this class for every message that we process ?
I believe we can declare a static readonly member that represents the JsonSerializerOptions that we'd use, and by doing so we apply some kind of fly-weight :)

src/Arcus.Messaging.Health/Arcus.Messaging.Health.csproj
```diff
@@ -28,7 +28,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="[8.0.13,10.0.0)" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    
+    <!-- TODO: will be removed in v3.0 -->
+    <PackageReference Include="System.Text.Json" Version="[8.0.5,10.0.0)" Condition="'$(TargetFramework)' == 'netstandard2.1'" />
```
fgheysels, line 33: same remark here on the version-range.
Also wondering why we check here if the $TargetFramework is netstandard, while in the other csproj file we check if the TargetFramework is not .net 8

src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
stijnmoreels: Think Microsoft uses major versions not in the normal sense, but more in line with their .NET versions. Think therefore that .NET 9 is reflected here, instead of being a breaking change version.

src/Arcus.Messaging.Health/Arcus.Messaging.Health.csproj
stijnmoreels: Ah yes, that we need to streamline.

src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
stijnmoreels: Bc I think that (just like from .NET 6 to .NET 8), there are breaking changes. Unlike from .NET 6 to .NET 7, believe.

fgheysels: Hmm, ok.
But then again, we only need this package when we target < .NET 8.
So, in that case, I think I would change the condition and specify that we only include this package when we're targetting .netstandard.
And in that case, I think it would be sufficient to specify that we want `Version=8.*` ?

Might be a bit clearer ?  Not that important anyway as we'll remove this in a later version

stijnmoreels: * We also need it (I saw) in .NET 6, as there is a missing feature for 'missing JSON properties' that is not available in the built-in version of .NET 6;
* And fine for me to do it instead with wildcards.

src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
```diff
@@ -27,9 +27,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!-- TODO: will be removed in v3.0 -->
     <PackageReference Include="Arcus.Observability.Telemetry.Serilog.Enrichers" Version="[3.0.0,4.0.0)" />
     <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.21.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="System.Text.Json" Version="[8.0.5,10.0.0)" Condition="'$(TargetFramework)' != 'net8.0'" />
```
stijnmoreels, line 33:
```suggestion
    <PackageReference Include="System.Text.Json" Version="8.*" Condition="'$(TargetFramework)' != 'net8.0'" />
```

src/Arcus.Messaging.Health/Arcus.Messaging.Health.csproj
```diff
@@ -28,7 +28,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="[8.0.13,10.0.0)" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    
+    <!-- TODO: will be removed in v3.0 -->
+    <PackageReference Include="System.Text.Json" Version="[8.0.5,10.0.0)" Condition="'$(TargetFramework)' != 'net8.0'" />
```
stijnmoreels, line 33:
```suggestion
    <PackageReference Include="System.Text.Json" Version="8.*" Condition="'$(TargetFramework)' != 'net8.0'" />
```

src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
fgheysels: And version 8.0.5 works for .NET 6 ?

fgheysels (approved):
I think it's OK, but I have just one final question :)

src/Arcus.Messaging.Abstractions/Arcus.Messaging.Abstractions.csproj
stijnmoreels: Ran the tests with both frameworks successfully, so I would guess so, yes.

stijnmoreels:
@netlify deploy-preview

fgheysels:
This one is OK for me, but it looks like there's one pending check (WIP) ?

PR #476, opened 2025-02-13
chore(remove): event-hubs pump project

stijnmoreels:
The Messaging library previously supported both Service bus and Event Hubs as message pumps. Due to maintainence costs, and the sheldom usage of message handling in event-sourcing systems, the maintainence cost for this is too high.

Related to the exercise described here #470 , this PR removes the Event Hubs message pump and related feature documentation. This can still be done in a minor release, as separate projects means separate NuGet packages. We only have to archive the NuGet package once we're finished with the release.


fgheysels (commented):
Just wondering if we shouldn't first deprecate before removing ?
We can have an offline chat on that.

Not necessary, I see since it's in a separate project we can just do that.

fgheysels:
I still think it's a pity to remove this pump though.
I wonder if the maintenance cost is that high ?  We might not use it very much atm, but I do see use-cases.  (Could've used it in the past in one of our projects f.i.)

fgheysels (approved):

PR #483, opened 2025-02-13
fix(service-bus): remove invalid flags attribute on topic subscription

stijnmoreels:
The `TopicSubscription` invalidly had the `[Flags]` attribute, which would indicate that the enumeration members could be used in combination with one another - which is not correct.

This PR removes the `[Flags]` attribute, making the intend of the enumeration more clear.


fgheysels (approved):
Nice catch!

PR #491, opened 2025-02-21
chore(serilog): deprecate serilog correlation enricher in abstractions

stijnmoreels:
The Serilog-functionality should not be hard-linked to our Messaging solution. Currently, there exists a Serilog enricher that enrichers log events with the current message correlation. This should be moved to a separate library in next versions.

This PR deprecates this Serilog enricher with a meaningful description.


src/Arcus.Messaging.Abstractions/Telemetry/MessageCorrelationInfoEnricher.cs
```diff
@@ -8,6 +8,7 @@ namespace Arcus.Messaging.Abstractions.Telemetry
     /// <summary>
     /// Logger enrichment of the <see cref="MessageCorrelationInfo" /> model.
     /// </summary>
+    [Obsolete("Will be moved in v3.0 outside the 'Abstractions' library in a separate Serilog-specific library, see the v3.0 migration guide for more information")]
```
fgheysels, line 11: I'd only do this once the alternative exists.

stijnmoreels: Yes, that's what the message says. Think this is also not often used directly.

fgheysels (approved):

fgheysels (approved):

PR #493, opened 2025-02-21
chore(style): add .editorconfig file proposal

stijnmoreels:
Adds an `.editorconfig` file (same as proposed to Arcus.Testing) to the solution.
This will help streamline this and future work on the project by us and contributors - which we can later also `dotnet format` verify during PR reviews.


fgheysels (approved):

stijnmoreels:
> Automatic style verification is nice! We can always fine-tune the settings once it's in use. The only thing I wonder is if we can't have this editor-config file in a separate repo, because now, it must be duplicated in all repos ?

Yes, but we already tried to do some centralizing like for YAML steps, and what I feel, is that for Arcus, we are simply with too little to have such an approach with multi-repo setups. Adds maintainence.

fgheysels:
I think this one can be completed but it looks like there's an issue with the CodeQL analyzer ?

PR #494, opened 2025-02-25
chore(sb): deprecate received message application property extension

stijnmoreels:
The entire `.ServiceBus.Core` project will be deprecated/removed, later. This PR deprecates the application property extension, as this is could be implicitly (pump project references the core project) be used by custom message handlers and deprecating it now will show that it will be removed later.


fgheysels:
Just did a quick search in our project codebase to see if we use it, but we do not.

`MessageHandler` implementations do not seem to have the possibility to directly reference the `ServiceBusReceivedMessage` object.   Message-properties are exposed via the `AzureServiceBusMessageContext`

stijnmoreels:
> Just did a quick search in our project codebase to see if we use it, but we do not.
> 
> `MessageHandler` implementations do not seem to have the possibility to directly reference the `ServiceBusReceivedMessage` object. Message-properties are exposed via the `AzureServiceBusMessageContext`

You could access them in fallback message handlers. But yes, not used by us.

fgheysels (approved):

PR #495, opened 2025-02-25
chore(sb): deprecate older rotate connection strings functionality

stijnmoreels:
Deprecates the functionality in the Service bus pump project that allows for rotation of conection strings. This used an older version of authentication, and is somewhat conflicting with the 'pump' project as it has little to do with message routing.

This PR deprecates all rotation-related functionality.


fgheysels:
I didn't even know this functionality existed in Arcus.Messaging.
TBH, I also think that this kind of functionality should not be the responsibility of Arcus.Messaging.  This is more an 'infra/devops' or 'operations-like' task.    I don't think you want a program that is responsible for consuming / processing messages be responsible for things like this.

IMHO, a good thing to have it removed.

fgheysels (approved):

PR #497, opened 2025-02-27
chore(deps): remove guard.net references from the pumps abstractions project

stijnmoreels:
The pumps abstractions still had some references to the Guard.NET library. This PR removes those references in favor of built-in functionlity - not using the .NET 8 functionlity just yet as this is not availble in .NET Standard 2.1.

### DIFF
--- a/src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
+++ b/src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Linq;
 using System.Threading.Tasks;
-using GuardNet;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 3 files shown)


src/Arcus.Messaging.Pumps.Abstractions/Resiliency/DefaultMessagePumpCircuitBreaker.cs
```diff
@@ -82,7 +85,10 @@ public MessagePumpCircuitState GetCircuitBreakerState(string jobId)
         /// <exception cref="InvalidOperationException">Thrown when not a single or more than one message pump could be found by the configured job ID.</exception>
         protected MessagePump GetRegisteredMessagePump(string jobId)
         {
-            Guard.NotNullOrWhitespace(jobId, nameof(jobId));
```
fgheysels, line 85: Since there were still references to the `Guard` class, this means that there must still be a reference to the `Guard.NET` project somewhere.  I believe we should find that reference and remove it from the csproj file as well.

stijnmoreels: It is a transient package reference from Arcus.Observability. Not something that we explicitly reference.

fgheysels: Ah Ok.  Thanks!

fgheysels (approved):

PR #500, opened 2025-02-27
chore(general): deprecate general fallback message handler

stijnmoreels:
Only concrete implementations of the message pump and message router will be available in the v3.0 space. This PR deprecates the general fallback message handler interface and registration so that it can be removed later on.

General message routing includes routing messages based on string instead of concrete types like ServiceBusReceivedMessage. The general message routing was required before when using it directly in Azure Functions, but since we removed the specific Azure Funcions support, we should remove this general message routing as well.

Relates to lightweight exercise described in https://github.com/arcus-azure/arcus.messaging/discussions/470

### DIFF
--- a/src/Arcus.Messaging.Abstractions/Extensions/MessageHandlerCollectionExtensions.cs
+++ b/src/Arcus.Messaging.Abstractions/Extensions/MessageHandlerCollectionExtensions.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <param name="collection">The collection of collection to use in the application.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="collection"/> is <c>null</c>.</exception>
         public static MessageHandlerCollection WithMessageHandler<TMessageHandler, TMessage, TMessageContext>(this MessageHandlerCollection collection)
-            where TMessageHandler : class, IMessageHandler<TMessage, TMessageContext> 
+            where TMessageHandler : class, IMessageHandler<TMessage, TMessageContext>
             where TMessage : class
             where TMessageContext : MessageContext
         {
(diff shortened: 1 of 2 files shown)


src/Arcus.Messaging.Abstractions/Extensions/MessageHandlerCollectionExtensions.cs
```diff
@@ -154,10 +157,11 @@ public static MessageHandlerCollection WithFallbackMessageHandler<TMessageHandle
         /// <param name="collection">The collection to add the fallback message handler to.</param>
         /// <param name="createImplementation">The function to create the fallback message handler.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="collection"/> or the <paramref name="createImplementation"/> is <c>null</c>.</exception>
+        [Obsolete("Will be removed in v3.0 as only concrete implementations of message handling will be supported from now on")]
```
fgheysels, line 160: So this means the complete concept of `FallbackMessageHandler`s will be removed ?

fgheysels: Ah, I see.....   Now I understand.
We have 'AzureServiceBusFallbackMessageHandler', and we have this 'general' ones that are actually never used atm.
Should someone ever wanted to create support for let's say RabbitMQ, these 'concepts' would need to be created specifically for RabbitMQ itself.

fgheysels (approved):

src/Arcus.Messaging.Abstractions/Extensions/MessageHandlerCollectionExtensions.cs
stijnmoreels: Yes, they are used implicitly in both the Service bus and EventHubs routing, but only there because we supported general message routing in the past. And since Service bus has a special case of message routing and we removed the EventHubs routing, there is no need to have a fully implemented general message routing available.

stijnmoreels: Can we resolve this @fgheysels ?

PR #513, opened 2025-03-13
chore(testing): use built-in arcus.testing temp env var

stijnmoreels:
Starting from Arcus.Testing v1.3, there exists a built-in temporary environment fixture that helps a lot with managing values and secrets throughout integration tests.

This PR removes our own local version in favor of the one provided by Arcus.Testing.


src/Arcus.Messaging.Tests.Integration/Arcus.Messaging.Tests.Integration.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="Arcus.Testing.Logging.Xunit" Version="1.1.0" />
     <PackageReference Include="Arcus.Testing.Security.Providers.InMemory" Version="1.1.0" />
     <PackageReference Include="Arcus.Testing.Storage.Blob" Version="1.2.0" />
+    <PackageReference Include="Arcus.Testing.Core" Version="1.3.0" />
```
stijnmoreels, line 17: Will eventually also be removed, as it will become a transient dependency once we use the Arcus.Testing package for Service bus.

fgheysels (approved):

PR #526, opened 2025-03-20
chore(observability): deprecate message correlation accessor

stijnmoreels:
As discussed in #470, the core messaging will solely focus on message handling/routing, which makes the hard-linked dependencies to Arcus.Security, Arcus.Observability, Serilog and Application Insights deprecated in the abstractions library. Previous PR's already deprecated Serilog and Arcus.Security functionality, like #491.

This PR deprecates the Arcus.Observability correlation accessor implementation, as this is a rather unnecessary implementation, since the correlation will already be provided via the message routing.

### DIFF
--- a/src/Arcus.Messaging.Abstractions/IMessageCorrelationInfoAccessor.cs
+++ b/src/Arcus.Messaging.Abstractions/IMessageCorrelationInfoAccessor.cs
@@ -1,4 +1,5 @@
-﻿using Arcus.Observability.Correlation;
+﻿using System;
+using Arcus.Observability.Correlation;
 
 namespace Arcus.Messaging.Abstractions
 {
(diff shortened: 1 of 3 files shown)


src/Arcus.Messaging.Abstractions/IMessageCorrelationInfoAccessor.cs
```diff
@@ -1,11 +1,13 @@
...
+﻿using System;
+using Arcus.Observability.Correlation;
 
 namespace Arcus.Messaging.Abstractions
 {
     /// <summary>
     /// Represents a marker interface for an <see cref="ICorrelationInfoAccessor{TCorrelationInfo}"/> implementation using the messaging <see cref="MessageCorrelationInfo"/>.
     /// </summary>
     /// <seealso cref="ICorrelationInfoAccessor{TCorrelationInfo}"/>
+    [Obsolete("Will be moved in v3.0 outside the 'Abstractions' library in a separate Serilog-specific library, see the v3.0 migration guide for more information")]
```
fgheysels, line 10: Is it possible to do this now already ?
As, imho, it makes no sense to deprecate a class now, without having the alternative ready ?
Or, we must make sure that the obsolete class is still part of the library when the alternative is thre

stijnmoreels: 1. If we leave the class non-deprecated, we have to wait for v4 to remove it;
2. The alternative will be available on v3, as it will include breaking changes to include an alternative.
As far as I see, there is no way of doing both removing all references (Serilog) and yet still provide support for all references.

fgheysels: Ok, fine for me.

fgheysels (approved):

PR #535, opened 2025-03-25
chore(message-context): provide additional service-bus properties via builder to message context

stijnmoreels:


src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
```diff
@@ -19,6 +20,7 @@ public class AzureServiceBusMessageContext : MessageContext
         /// <param name="properties">The contextual properties provided on the message provided by the message publisher.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="systemProperties"/> is <c>null</c>.</exception>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="jobId"/> is blank.</exception>
+        [Obsolete("Use the " + nameof(ServiceBusMessageContextBuilder) + " to construct message contexts")]
```
fgheysels, line 23: As mentioned in our meeting earlier today, I'm not convinced that this is a better approach.
You need a min number of required properties that must be set; those required properties must be set, be it via the ctor or via a Builder pattern.  Using a Builder doesn't resolve the fact that you can pass in less properties into the ctor or `Create` method of the Builder.

For instance the `jobId` ; isn't that a required property ?

stijnmoreels: Yes, only currently there can be message contexts created without receiver and job ID. So, if we decide that these are indeed always required, then we might need yo deprecate some more. Maybe even the constructors of general message contexts. 🤔

fgheysels: And how is that achieved now ?  Because the constructor accepts the jobId ?

stijnmoreels: Yes, there are different constructor overloads, with/without additional info.

PR #541, opened 2025-04-01
feat(correlation): deprecate message correlation result in favor of message operation result

stijnmoreels:
As described in the discussion #470, the message corelation system should be extracted from the core messaging functionality to support multiple systems.
This PR deprecates the `MessageCorrelationResult` (as it was incorrectly hard-linked to the Serilog/AppInsights correlation system), in favor of a mor general `MessageOperationResult` which will get used when tracking requests in message pumps.

### DIFF
--- a/src/Arcus.Messaging.Abstractions/MessageCorrelationResult.cs
+++ b/src/Arcus.Messaging.Abstractions/MessageCorrelationResult.cs
@@ -10,6 +10,7 @@ namespace Arcus.Messaging.Abstractions
     /// Represents the correlation result of a received Azure Service Bus message.
     /// This result will act as the scope of the request telemetry.
     /// </summary>
+    [Obsolete("Will be removed in v3.0 in favor of " + nameof(MessageOperationResult))]
     public sealed class MessageCorrelationResult : IDisposable
     {
         private readonly TelemetryClient _telemetryClient;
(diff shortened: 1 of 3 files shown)


src/Arcus.Messaging.Abstractions/Telemetry/MessageOperationResult.cs
```diff
@@ -0,0 +1,67 @@
...
+            StopOperation(IsSuccessful, _startTime, _watch.Elapsed);
+        }
+
+        /// <summary>
+        /// Finalizes the tracked operation in the concrete telemetry system, based on the operation results.
+        /// </summary>
+        /// <param name="isSuccessful">The boolean flag to indicate whether the operation was successful.</param>
+        /// <param name="startTime">The date when the operation started.</param>
+        /// <param name="duration">The time it took for the operation to run.</param>
+        protected virtual void StopOperation(bool isSuccessful, DateTimeOffset startTime, TimeSpan duration)
```
fgheysels, line 63: I see that the constructor is protected.  In that case, why is this method not defined as   `abstract` (and the class as abstract as well)?

stijnmoreels: Its a bit double. Wanted to both think about future solutions, as well as the current ones. Maybe having two custom classes for each correlation solution is not very dev-friendly.

In this case: yeah, maybe we don't need to worry about this just now, and mark them both `abstract`.

stijnmoreels: Done that here: https://github.com/arcus-azure/arcus.messaging/pull/541/commits/02c37c7216cb4cf009f0e86818f81de60be843dd

fgheysels (approved):

PR #542, opened 2025-04-01
feat(az.sb.telemetry): add telemetry options alternative in message pump options

stijnmoreels:
As described in discussion #470, the message correlation system choice should be extracted from the core messaging functionality. The plan is to expose an interface that can be implemented by specific message correlation systems. Interacting with this interface will be done in the message pump, therefore, the telemetry options to manipulate the message correlation should also be in the message pump options.

This PR creates a new `.Telemetry` options property in the Azure Service bus message options, which should be used instead of the (now) deprecated `.Routing.Telemetry` options. Making it more clear, correct and maintainable.

### DIFF
--- a/src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptions.cs
+++ b/src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Arcus.Messaging.Abstractions.MessageHandling;
 using Arcus.Messaging.Abstractions.ServiceBus.MessageHandling;
 
 #pragma warning disable CS0618
@@ -146,6 +147,11 @@ namespace Arcus.Messaging.Pumps.ServiceBus.Configuration
         /// Gets the consumer-configurable options to change the behavior of the message router.
         /// </summary>
         public AzureServiceBusMessageRouterOptions Routing { get; }
+
+        /// <summary>
+        /// Gets the consumer configurable options model to change the behavior of the tracked Azure Service bus request telemetry.
+        /// </summary>
+        public MessageTelemetryOptions Telemetry => Routing.Telemetry;
 
         /// <summary>
         /// Gets the default consumer-configurable options for Azure Service Bus Queue message pumps.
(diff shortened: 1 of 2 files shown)


src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -147,6 +148,11 @@ public int MaximumUnauthorizedExceptionsBeforeRestart
         /// </summary>
         public AzureServiceBusMessageRouterOptions Routing { get; }
 
+        /// <summary>
+        /// Gets the consumer configurable options model to change the behavior of the tracked Azure Service bus request telemetry.
+        /// </summary>
+        public MessageTelemetryOptions Telemetry => Routing.Telemetry;
+
```
fgheysels, line 155: I don't see why we need this property ?
These `MessageTelemetryOptions` are already available via the `Routing` property.  Byt maybe I 'm missing some context here ?

stijnmoreels: The `Routing.Telemetry` property is marked deprecated, it is moved to the pump options, as it is the pump solely handles telemetry. It is also an exercise in making the `MessageRouterOptions` simpler - and in the future possibly to remove it, too.

fgheysels: Ok, but if `Routing.Telemetry` is marked as deprecated, doesn't the compiler emit a warning as you're calling an obsolete property now on L154 ?

stijnmoreels: Deprecation warnings have been temporary disabled on this file, see L5: `#pragma warning disable CS0618`.

fgheysels: Ok, haven't seen that.  But this means that, at some point in time the implementation of the `Telemetry` property will have to be changed ?

stijnmoreels: Yes, when we're working on v3, we can start removing stuff and move the get/set logic a level higher.

PR #545, opened 2025-04-01
docs(migration.v3.0): add new service-bus message pump registration to v3.0 migration guide

stijnmoreels:
In a previous phase, the Azure Service bus message pump registration was reviewed and simplified to handle a dependency-less approach without overload the user with overloads.

This PR adds this new way of registering the message pump to the v3.0 migration guide.
Relates to #506

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -61,7 +61,7 @@ module.exports = {
     prism: {
       theme: lightCodeTheme,
       darkTheme: darkCodeTheme,
-      additionalLanguages: ['csharp'],
+      additionalLanguages: ['csharp', 'diff'],
     },
   },
   presets: [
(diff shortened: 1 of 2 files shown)


docs/docusaurus.config.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
     prism: {
       theme: lightCodeTheme,
       darkTheme: darkCodeTheme,
-      additionalLanguages: ['csharp'],
+      additionalLanguages: ['csharp', 'diff'],
```
stijnmoreels, line 64:
```suggestion
      additionalLanguages: ['csharp', 'powershell', 'diff'],
```

fgheysels (approved):
Migration guides will make life easier :)

Issue #547, opened 2025-04-04
Possibility to deadletter or abandon a message via a methodcall on AzureServiceBusMessageContext


fgheysels:
We can currently easily deadletter or abandon a message from inside an `AzureServiceBusMessageHandler` implementation.
However, when developing you sometimes instantiate other classes and perform further processing of the message there.   In some situations though, you need to abandon or deadletter a message.
It would be good if we could easily do this by calling a method on the AzureServiceBusMessageContext instead of having to call the methods that make that happen which are member methods of the `AzureServiceBusMessageHandler` class.

stijnmoreels:
Could make the message routing also a lot simpler. I believe that we could mark the `AzureServiceBusMessageHandler` also deprecated, in that case, in favor of methods on the context.

stijnmoreels:
Wondering if the 'fallback' functionality has any added-value if we fully expose the Service bus-specific operations via the context.

fgheysels:
To be honest, I never really understood the 'fallback' functionality.  If you need to 'fallback', then I think you (as a user of Arcus.Messaging) really missed implementing some functionality / feature ?

Next to that, wouldn't deprecating `AzureServiceBusMessageHandler ` a huge breaking change ?

stijnmoreels:
1. the 'fallback' functionality was (I think) implemented to run custom code in case of the message handlers not being matched. Think it is mostely used in scenarios where the type of messages is not always known, too?
2. no, not so much a breaking change. Since this is an additional feature, and especially if we decide to move to the message context, it might be better to deprecate the `AzureServiceBusMessageHandler` template class (and maybe the entire 'fallback' functionalty with it?).

lightpeak:
I would also like to ask to look at how to set the **deadLetterErrorDescription** and **deadLetterReason** when using Function Apps.

I was unsuccessful because for some reason Arcus kept pushing the default error that 'no suitable handler was found'.

With the latest official servicebus package you can do this out of the box like this:
```
catch (Exception ex)
{
    await messageActions.DeadLetterMessageAsync(message, ex.GetType().ToString(), ex.Message);
   
}
```

But back to Arcus. It was suggested to inject the **servicebusreceiver** manually but this also resulted in an error, because the message actually came in through the function app trigger (with it's own receiver, which had a lock on it) and the servicebusreceiver we initiated was from a different "scope" which as a consequence had no access to the message.

Some guidance would be appreciated!

Thank you and kind regards,
Jelle

[...]

PR #551, opened 2025-04-11
feat(sb): move service-bus operations to message-context

stijnmoreels:
Moves the Azure Service bus-specific operations (Complete, Abandon, Dead-Letter) to the message context, so that the following functionality can be made deprecated:
* Azure Service bus message template classes for regular/fallback message handlers, as these were solely used to trigger the Azure Service bus operations;
* General fallback functionality, as this is now also unnecessary for any of the 'messaging' implementations;

Relates to #547

### DIFF
--- a/src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
+++ b/src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
@@ -160,6 +160,18 @@ namespace Arcus.Messaging.Abstractions.ServiceBus
         }
 
         /// <summary>
+        /// Dead letters the Azure Service Bus message on Azure with a reason why the message needs to be dead lettered.
+        /// </summary>
+        /// <param name="deadLetterReason">The reason why the message should be dead lettered.</param>
+        /// <param name="deadLetterErrorDescription">The optional extra description of the dead letter error.</param>
+        /// <param name="cancellationToken">The optional <see cref="CancellationToken" /> instance to signal the request to cancel the operation.</param>
+        /// <exception cref="InvalidOperationException">Thrown when the message handler was not initialized correctly.</exception>
+        public async Task DeadLetterMessageAsync(string deadLetterReason, string deadLetterErrorDescription, CancellationToken cancellationToken)
+        {
+            await DeadLetterMessageAsync(deadLetterReason, deadLetterErrorDescription, newMessageProperties: null, cancellationToken);
+        }
+
+        /// <summary>
         /// Dead letters the Azure Service Bus message on Azure while providing <paramref name="newMessageProperties"/> for properties that has to be modified in the process.
         /// </summary>
         /// <param name="deadLetterReason">The reason why the message should be dead lettered.</param>
@@ -172,20 +184,6 @@ namespace Arcus.Messaging.Abstractions.ServiceBus
             EnsureServiceBusFields();
 
             await _receiver.DeadLetterMessageAsync(_message, newMessageProperties, deadLetterReason, deadLetterErrorDescription, cancellationToken);
-        }
-
-        /// <summary>
-        /// Dead letters the Azure Service Bus message on Azure with a reason why the message needs to be dead lettered.
-        /// </summary>
-        /// <param name="deadLetterReason">The reason why the message should be dead lettered.</param>
-        /// <param name="deadLetterErrorDescription">The optional extra description of the dead letter error.</param>
-        /// <param name="cancellationToken">The optional <see cref="CancellationToken" /> instance to signal the request to cancel the operation.</param>
-        /// <exception cref="InvalidOperationException">Thrown when the message handler was not initialized correctly.</exception>
-        public async Task DeadLetterMessageAsync(string deadLetterReason, string deadLetterErrorDescription, CancellationToken cancellationToken)
-        {
-            EnsureServiceBusFields();
-
-            await _receiver.DeadLetterMessageAsync(_message, deadLetterReason, deadLetterErrorDescription, cancellationToken);
         }
 
         /// <summary>
--- a/src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
+++ b/src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using Azure.Messaging.ServiceBus;
 using Microsoft.Extensions.Logging;
 
@@ -11,6 +13,9 @@ namespace Arcus.Messaging.Abstractions.ServiceBus
     /// </summary>
     public class AzureServiceBusMessageContext : MessageContext
     {
+        private readonly ServiceBusReceiver _receiver;
+        private readonly ServiceBusReceivedMessage _message;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="AzureServiceBusMessageContext"/> class.
         /// </summary>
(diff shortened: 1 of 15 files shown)


stijnmoreels:
Waiting a bit on #268 to update the Service bus-specific operations in the feature docs.

fgheysels:
> Waiting a bit on #268 to update the Service bus-specific operations in the feature docs.

I don't understand what you mean here :)  You're linking to a PR that has been merged ?

src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
```diff
@@ -139,5 +147,70 @@ public static AzureServiceBusMessageContext Create(
 
             return new AzureServiceBusMessageContext(jobId, entityType, receiver, message);
         }
+
+        /// <summary>
+        /// Completes the Azure Service Bus message on Azure. This will delete the message from the service.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when the message handler was not initialized yet.</exception>
+        public async Task CompleteMessageAsync(CancellationToken cancellationToken)
```
fgheysels, line 155: Shouldn't these methods be declared as virtual or abstract methods on the `MessageContext` instead ?
I think these describe common functionality, whether you're working with Azure ServiceBus or any other message-broker, they describe some common messaging concepts imho.   Specific MessageContext implementations can then override / implement these methods.

I'm also thinking in terms of testability.  Suppose I want to create a unit-test for a certain MessageHandler, and I want to verify if the `DeadLetterMessageAsync` method is called in a certain situation, will I be able to do that when this is a 'concrete' method.  Can I easily mock/stub this ?

src/Arcus.Messaging.Abstractions.ServiceBus/Extensions/ServiceBusMessageHandlerExtensions.cs
```diff
@@ -109,6 +109,7 @@ public static ServiceBusMessageHandlerCollection WithServiceBusMessageHandler<TM
         /// <typeparam name="TMessageHandler">The type of the fallback message handler.</typeparam>
         /// <param name="handlers">The handlers to add the fallback message handler to.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="handlers"/> is <c>null</c>.</exception>
+        [Obsolete("Will be removed in v3.0, please use the Azure service bus operations on the " + nameof(AzureServiceBusMessageContext) + " instead of defining fallback message handlers")]
         public static ServiceBusMessageHandlerCollection WithServiceBusFallbackMessageHandler<TMessageHandler>(
```
fgheysels, line 113: It's not entirely clear to me how the Complete/Abandon/Deadletter methods on the `AzureServiceBusMessageContext` will replace the fallback messagehandlers ?  (Might be because I never ever used the Fallback mechanism).

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageHandler.cs
```diff
@@ -9,6 +9,7 @@ namespace Arcus.Messaging.Abstractions.ServiceBus.MessageHandling
     /// <summary>
     /// Represents a <see cref="IAzureServiceBusMessageHandler{TMessage}"/> template to control the Azure Service Bus message operations during the handling of the deserialized message.
     /// </summary>
+    [Obsolete("Will be removed in v3.0, please use the Azure service bus operations on the " + nameof(AzureServiceBusMessageContext) + " instead")]
```
fgheysels, line 12: Shouldn't the deprecation description indicate that the `IAzureServiceBusMessageHandler` interface must be used instead ?

src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouter.cs
```diff
@@ -263,6 +263,7 @@ protected async Task RouteMessageAsync<TMessageContext>(
             }
         }
 
+        [Obsolete("Will be removed in v3.0, as the entire fallback message handler structure becomes unnecessary")]
         private async Task<bool> TryProcessMessageAsync<TMessageContext>(
```
fgheysels, line 267: Sure that this method is obsolete as well ?  Given the name, I wouldn't suspect that it has something to do with the fallback mechanism ?

src/Arcus.Messaging.Abstractions.ServiceBus/Extensions/ServiceBusMessageHandlerExtensions.cs
stijnmoreels: The fallback message handlers were usually used to call Service bus-operations. Like, when you want to have a custom dead-letter message in case none of the message handlers match. To simplify the messaging system, removing the fallback handlers will also make it more easy/flexible for consumers to use the Service bus operations, which having to implement fallback message handlers.

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageHandler.cs
stijnmoreels: Ah yes, we can add that as well. Thx! 👍

src/Arcus.Messaging.Abstractions/MessageHandling/MessageRouter.cs
stijnmoreels: The callee of this method is already marked deprecated. This additional deprecation was made since this method is calling fallback functionality in its body.

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageHandler.cs
stijnmoreels: Done that here: https://github.com/arcus-azure/arcus.messaging/pull/551/commits/ae51309029e63e8b3e78c93d1e7c0a9463741b03

src/Arcus.Messaging.Abstractions.ServiceBus/Extensions/ServiceBusMessageHandlerExtensions.cs
fgheysels: Ah, if that was the idea of the fallback handlers, I see :)  But, IMHO we've improved / simplified this a bit by making sure that the MessageRouter (I think) is now deadlettering the message itself if no suitable MessageHandler can be found.

src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
fgheysels: I see.  One could argue that in those cases , the implementation of these methods should throw a NotSupportedException.  But, since we currently solely support Azure Service Bus, we can also consider this as a YAGNI atm.

For testing purposes, I like to use Moq and should we have an interface that defines these methods, it's easier -imho- to use that in tests instead of having to provide in-mem ServiceBusReceivers  etc

stijnmoreels: Since these are `virtual`, `Moq`, could be used. But, for complexer scenario's, with actual Mock-verification, I would always recommend a dedicated implementation. 🙂

src/Arcus.Messaging.Abstractions.ServiceBus/Extensions/ServiceBusMessageHandlerExtensions.cs
stijnmoreels: Yes, indeed.

fgheysels (approved):

src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
stijnmoreels: Can we resolve this conv, or do we need to discus something else?

stijnmoreels: I meant on the `ServiceBusReceiver`. It also has a empty `protrcted` constructor just fir this testing purpose https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusreceiver.completemessageasync?view=azure-dotnet

fgheysels: I can live with it for now.

fgheysels (approved):
Might require some refactoring for Arcus users, but it's an improvement imho :)

PR #546, opened 2025-04-04
docs(az.service-bus): restructure az service-bus messaging

stijnmoreels:
As described in the discussion #470 , there is a lot of functionality changed/removed/enhanced. Because of this exercise, it was a good time to also review the current Azure Service bus messaging feature documentation.

This PR simplifies and restructures the Azure Service bus messaging, by splitting the 'getting started' from the advanced 'customization' features. Making the documentation much more accessible.

PR #557, opened 2025-05-06
chore(general): deprecate message pump lifetime restartable functionality

stijnmoreels:
As first mentioned in the discussion #470 and re-confirmed in #546 , the `IMessagePumpLifetime`/`IRestarableMessagePump` are deprecated since their initial purpose was to be able to provide a system to 'pause' the message pump when the external dependency system can't keep up - which is exactly what the new circuit breaker functionality provides.


stijnmoreels:
We then might also want to remove the `Options.MaximumUnauthorizedExceptionsBeforeRestart` functionality, as it is using this 'restart' functionality. Since the message pump has now been registered with a custom `ServiceBusClient` registration, the authentication has also been moved outside of the message pump's responsibility.

fgheysels (approved):

PR #565, opened 2025-05-22
chore(pipelines): reduce duplication by extracting tests as job templates

stijnmoreels:
This PR extracts both the unit & integration tests job runs in separate job templates. This allows for less duplication between the CI build and the NuGet release pipeline, as they now both towards the same job template.

By reducing the duplication and having a direct reference, changes to one pipeline immediately effects the other - which in this case, is expected/best-practice as we want. It also helps with any possible failures/missing code in the NuGet release, which would only be flagged upon actually releasing.

This approach was already taken in Arcus Testing, but is now also introduced in Messaging.


fgheysels:
Nice approach!

PR #576, opened 2025-05-27
chore(general): remove deprecated general 'route message' with string

stijnmoreels:
> 👉 Since we're working on v3.0, we can start removing stuff for real.

This PR removes the deprecated `IMessageRouter.RouteMessageAsync` method that takes in a `string` message in favor of concrete type implementations in message pump/router solutions.


src/Arcus.Messaging.Abstractions/MessageHandling/IMessageRouter.cs
```diff
@@ -10,24 +8,5 @@ namespace Arcus.Messaging.Abstractions.MessageHandling
     [Obsolete("Will be removed in v3.0 as only concrete implementations of message routing will be supported from now on")]
     public interface IMessageRouter
```
fgheysels, line 9: Why don't we remove the interface entirely ?

stijnmoreels: Step by step, keeping changes separated to reduce ripple effects.

Issue #484, opened 2025-02-13
🗑️ [Remove]: deprecated functionality in v3.0


stijnmoreels:
Starting from v3.0, there is a lot of deprecated functionality that needs to be removed. This issue acts as a parent for all the pending work.

### Arcus.Messaging.Abstractions.ServiceBus
- [x] Remove deprecated `AddServiceBusMessageRouting` extension #611 
- [x] Remove deprecated `WithServiceBusMessageHandler` extensions with message body filter #590 
- [x] Remove deprecated `WithServiceBusMessageHandler` extensions with message context filter #589 
- [x] Remove deprecated `WithServiceBusMessageHandler` with serializer #566
- [x] Remove deprecated `WithServiceBusMessageHandler` with message body filter + serializer #566
- [x] Remove deprecated `WithServiceBusMessageHandler` extensions with message context filter + serializer #598 
- [x] Remove deprecated `WithServiceBusMessageHandler` extensions with message context/body filter + serializer #598 
- [x] Remove deprecated `WithServiceBusFallbackMessageHandler` extensions #572 
- [x] Remove deprecated `AzureServiceBusFallbackMessageHandler` #593 
- [x] Remove deprecated `AzureServiceBusMessageHandler` #574
- [x] Remove deprecated `AzureServiceBusMessageHandlerTemplate` #593 
- [x] Remove deprecated `AzureServiceBusMessageRouter` constructors #607 
- [x] Remove deprecated `IAzureServiceBusFallbackMessageHandler` #599 
- [x] Remove deprecated `IAzureServiceBusMessageRouter` methods #588 
- [x] Remove deprecated `AzureServiceBusMessageContext` constructors #575
- [x] Remove deprecated `GetMessageContext` extensions #575
- [x] Remove deprecated `GetSystemProperties` extension #598 
- [x] Remove deprecated `GetServiceBusReceiver` extension #570
- [x] Remove .NET 6 + .NET Standard 2.1 #578 

### Arcus.Messaging.Pumps.ServiceBus
- [x] Remove deprecated `AzureServiceBusCorrelationOptions` #600 
- [x] Remove deprecated `IAzureServiceBusQueueMessagePumpOptions` #600 
- [x] Remove deprecated `IAzureServiceBusQueueMessagePumpOptions` #600 
- [x] Make `AzureServiceBusMessagePumpSettings` internal #584 
- [x] Remove deprecated `AddServiceBusQueueMessagePump` extensions #591 
[...]

PR #593, opened 2025-06-13
chore(az.sb): remove deprecated fallback from az sb msg router

stijnmoreels:
> 👉 Since we're working on v3.0, we can remove stuff for real.

This PR removes the usage of 'fallback message handlers' in the Azure Service Bus message router. This means that the specific Azure Service Bus message handler template classes can now also be removed.

Relates to #484


src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
```diff
@@ -272,40 +266,14 @@ private async Task<MessageProcessingResult> TryRouteMessageWithPotentialFallback
                     }
```
fgheysels, line 266: Just wondering; where do we call 'Complete' if the message was succesfully processed ?
In this method, I see calls to Abandon or to Deadletter the message, depending on the case, but I don't see a call to Complete the message, when the message was processed succesfully.
Probably done somewhere else, but maybe it makes the code more readable / understandable if Complete is called here as well ?

src/Arcus.Messaging.Tests.Workers.ServiceBus/MessageHandlers/DeadLetterAzureServiceMessageHandler.cs
```diff
@@ -13,7 +13,7 @@ public class DeadLetterAzureServiceMessageHandler : IAzureServiceBusMessageHandl
         private readonly ILogger<DeadLetterAzureServiceMessageHandler> _logger;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="AzureServiceBusMessageHandlerTemplate"/> class.
+        /// Initializes a new instance of the <see cref="DeadLetterAzureServiceMessageHandler"/> class.
```
fgheysels, line 16: Eagle eyes :D

fgheysels (approved):

Issue #489, opened 2025-02-14
Support Session-enabled topics and queues in Arcus.Messaging


fgheysels:
It is currently not possible to consume a session-enabled topic or queue using Arcus.Messaging.
This is a feature which might not be so often used, but is definitely useful in some scenario's, therefore I think it might be useful to support it.

Unfortunately, this won't be an easy task to do, as the .NET SDK exposes a specific `ServiceBusReceiver` to consume session-enabled entities: the [ServiceBusSessionReceiver class](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebussessionreceiver?view=azure-dotnet).

I believe that this means that we need to define a new message-pump type in Arcus.Messaging.  The thing that makes it difficult, is that it seems that the `ServiceBusSessionReceiver` is only able to receive messages from one particular session.  This means that we would need to have multiple `ServiceBusSessionReceiver` instances per message-pump.  See also [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/samples/Sample03_SendReceiveSessions.md).

I've been looking at how the Azure Function bindings would cope with this, and I've seen that there exists a setting called [`MaxConcurrentSessions`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.servicebus.servicebusoptions.maxconcurrentsessions?view=azure-dotnet) which controls the amount of sessions (ServiceBusSessionReceiver instances?) that can be active at once. 

More background info on ServiceBus and sessions can be found [here](https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-sessions)

fgheysels:
I've been thinking about this ....

Would it be an idea to introduce an `AzureServiceBusSessionMessagePump`.  This class would make use of the `ServiceBusSessionProcessor` class which makes it easy to work with ServiceBus sessions.
Since working with ServiceBus sessions means that some additional settings must be configured, we'd also need an additional `AzureServiceBusSessionsMessagePumpSettings` class.

Drawback of this approach is maybe that it would be difficult (or impossible) to support CircuitBreaker functionality.  However, CircuitBreaker functionality on a MessagePump that works with ServiceBus sessions can be left out of scope for now, as this is functionality which might be seldom used in a scenario where ServiceBus sessions are being used.

PR #597, opened 2025-06-23
feat: Add ServiceBus MessagePump with session support

fgheysels:
Add an AzureServiceBusSessionMessagePump which can be used to read messages from Azure ServiceBus with taking Sessions into consideration.

This is a first and simple implementation.  Note that this message-pump doesn't have circuitbreaker support.

Relates to #489


fgheysels:
Created this PR as a spike.  
We need this behavior on a project, and I wanted to see / test what is possible.

src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
```diff
@@ -103,7 +110,7 @@ public static AzureServiceBusMessageContext Create(
         /// Completes the Azure Service Bus message on Azure. This will delete the message from the service.
         /// </summary>
         /// <exception cref="InvalidOperationException">Thrown when the message handler was not initialized yet.</exception>
-        public async Task CompleteMessageAsync(CancellationToken cancellationToken)
+        public virtual async Task CompleteMessageAsync(CancellationToken cancellationToken)
```
fgheysels, line 113: Defined these methods as `virtual` since I've created an `AzureServiceBusSessionMessageContext` class that inherits from this class.
The overriden methods use the `CompleteAsync` etc.. methods that are members of the `ProcessSessionMessageEventArgs` class.  That's necessary because the inherited class doesn't contain or has no access to a `ServiceBusReceiver` instance

fgheysels: Actually, I think it might be better to define an `IAzureServiceBusMessageContext` interface.
Both message-context implementations could then implement that interface.  The `IAzureServiceBusMessageHandler` implementations would then receive an instance of that interface.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
```diff
@@ -127,7 +127,6 @@ private static ProcessMessageAsync DetermineMessageImplementation<TMessage, TMes
             return async (rawMessage, generalMessageContext, correlationInfo, cancellationToken) =>
             {
                 if (rawMessage is TMessage message
-                    && generalMessageContext.GetType() == typeof(TMessageContext)
```
fgheysels, line 130: I removed this additional check because it's redundant (the check on L131 does the same) + this check caused the condition to evaluate to `false` when using the `AzureServiceBusSessionMessageContext`.
Although `AzureServiceBusSessionManagementMessageContext` is a `AzureServiceBusManagementMessageContext`, this comparison returned `false` because it was not exactly the same type.  (`IsAssignableFrom` would've been a better choice, but given the fact that we do a similar check on L131, I decided to just remove this check).

fgheysels:
> Ok, I've marked it as a 'draft' then, as to not to get confused.

I've published the PR right now as I think it would be good to get a first review.

src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
fgheysels: I'd introduce an interface to make sure that we only expose the 'common' properties and functionality.

An additional class for the `AzureServiceBusSessionMessagePump` is required because:
- it exposes the `SessionId` property, which is not relevant in a non-session scenario.  However, technically it could be present there as well as you may have a SessionId when not using sessions, in that case, it's just being ignored.
- More importantly, the implementation of completing / deadlettering / .... a message is different for the regular message-pump and the session-message-pump.   Completing a message has to be done via the `messageReceiver` in the `AzureServiceBusMessagePump`.  The `AzureServiceBusSessionMessagePump` doesn't have a `messageReceiver`.  Instead, the `ServiceBusSessionProcessor` is being used.  (If we wouldn't use the `ServiceBusSessionProcessor`, we'd have to keep track of multiple `ServiceBusSessionReceiver` instances.

So, in fact the goal of introducing that interface would actually be to hide the technicalities of how messages are completed / deadlettered / etc...

> Believe that instead of subclassing, we should have different ways to create a message context. Same for the message pump ➡️ user/consumer shouldn't concern themselves on internal details on 'how' messages are received/settled.

Always eager to learn the ideas you might have.  However, keep in mind that using a regular 'messagepump' or using a 'session aware' messagepump should be a considerate decision that must be taken by the developer.  Both have their own set of configuration-properties that must be set, and when using a 'session ware' messagepump, you have to use a topic-subscription (or queue) where session-support is explicitly enabled.

stijnmoreels: <details>
<summary>An example of how the message context could look like, without breaking changes or extra public classes:</summary>

```csharp
using System;
using System.Collections.Generic;
using System.Linq;
using System.Threading;
using System.Threading.Tasks;
using Azure.Messaging.ServiceBus;
using Microsoft.Extensions.Logging;

namespace Arcus.Messaging.Abstractions.ServiceBus
{
    /// <summary>
    /// Represents the contextual information concerning an Azure Service Bus message.
    /// </summary>
    public class AzureServiceBusMessageContext : MessageContext
    {
        private readonly IMessageSettleStrategy _messageSettle;

        /// <summary>
        /// Initializes a new instance of the <see cref="AzureServiceBusMessageContext"/> class.
        /// </summary>
        private AzureServiceBusMessageContext(
            string jobId,
            string entityPath,
            string fullyQualifiedNamespace,
            ServiceBusEntityType entityType,
            ServiceBusReceivedMessage message,
            IMessageSettleStrategy messageSettle)
            : base(message.MessageId, jobId, message.ApplicationProperties.ToDictionary(item => item.Key, item => item.Value))
        {
            _messageSettle = messageSettle;

            FullyQualifiedNamespace = fullyQualifiedNamespace;
            EntityPath = entityPath;
            EntityType = entityType;
            SystemProperties = AzureServiceBusSystemProperties.CreateFrom(message);
            LockToken = message.LockToken;
            DeliveryCount = message.DeliveryCount;
        }

        /// <summary>
        /// Gets the fully qualified Azure Service bus namespace that the message pump is associated with.
        /// This is likely to be similar to <c>{yournamespace}.servicebus.windows.net</c>.
        /// </summary>
        public string FullyQualifiedNamespace { get; init; }

        /// <summary>
        /// Gets the path of the Azure Service bus entity that the message pump is connected to,
        /// specific to the Azure Service bus namespace that contains it.
        /// </summary>
        public string EntityPath { get; init; }

        /// <summary>
        /// Gets the type of the Azure Service Bus entity on which the message was received.
        /// </summary>
        public ServiceBusEntityType EntityType { get; }

        /// <summary>
        /// Gets the contextual properties provided on the message provided by the Azure Service Bus runtime
        /// </summary>
        public AzureServiceBusSystemProperties SystemProperties { get; }

        /// <summary>
        /// Gets the token used to lock an individual message for processing
        /// </summary>
        public string LockToken { get; }

        /// <summary>
        /// Gets the amount of times a message was delivered
        /// </summary>
        /// <remarks>This increases when a message is abandoned and re-delivered for processing</remarks>
        public int DeliveryCount { get; }

        /// <summary>
        /// Creates a new instance of the <see cref="AzureServiceBusMessageContext"/> based on the current Azure Service bus situation.
        /// </summary>
        /// <param name="jobId">The unique ID to identity the Azure Service bus message pump that is responsible for pumping messages from the <paramref name="receiver"/>.</param>
        /// <param name="entityType">The type of Azure Service bus entity that the <paramref name="receiver"/> receives from.</param>
        /// <param name="receiver">The Azure Service bus receiver that is responsible for receiving the <paramref name="message"/>.</param>
        /// <param name="message">The Azure Service bus message that is currently being processed.</param>
        /// <exception cref="ArgumentNullException">Thrown when one of the parameters is <c>null</c>.</exception>
        public static AzureServiceBusMessageContext Create(
            string jobId,
            ServiceBusEntityType entityType,
            ServiceBusReceiver receiver,
            ServiceBusReceivedMessage message)
        {
            ArgumentException.ThrowIfNullOrWhiteSpace(jobId);
            ArgumentNullException.ThrowIfNull(receiver);
            ArgumentNullException.ThrowIfNull(message);

            var messageSettle = new MessageSettleViaReceiver(receiver, message);
            return new AzureServiceBusMessageContext(
                jobId,
                receiver.EntityPath,
                receiver.FullyQualifiedNamespace,
                entityType,
                message,
                messageSettle);
        }

        /// <summary>
        /// Creates a new instance of the <see cref="AzureServiceBusMessageContext"/> based on the current Azure Service bus situation.
        /// </summary>
        /// <param name="jobId">The unique ID to identity the Azure Service bus message pump that is responsible for pumping the message <paramref name="messageArgs"/>.</param>
        /// <param name="entityType">The type of Azure Service bus entity that the <paramref name="messageArgs"/> got received from.</param>
        /// <param name="messageArgs">The received args for the current session-aware message pump.</param>
        public static AzureServiceBusMessageContext Create(
            string jobId,
            ServiceBusEntityType entityType,
            ProcessSessionMessageEventArgs messageArgs)
        {
            ArgumentException.ThrowIfNullOrWhiteSpace(jobId);
            ArgumentNullException.ThrowIfNull(messageArgs);

            var messageSettle = new MessageSettleViaEventArgs(messageArgs);
            return new AzureServiceBusMessageContext(
                jobId,
                messageArgs.EntityPath,
                messageArgs.FullyQualifiedNamespace,
                entityType,
                messageArgs.Message,
                messageSettle);
        }

        /// <summary>
        /// Completes the Azure Service Bus message on Azure. This will delete the message from the service.
        /// </summary>
        /// <exception cref="InvalidOperationException">Thrown when the message handler was not initialized yet.</exception>
        public async Task CompleteMessageAsync(CancellationToken cancellationToken)
        {
            await _messageSettle.CompleteMessageAsync(cancellationToken);
        }

        /// <summary>
        /// Dead letters the Azure Service Bus message on Azure with a reason why the message needs to be dead lettered.
        /// </summary>
        /// <param name="deadLetterReason">The reason why the message should be dead lettered.</param>
        /// <param name="deadLetterErrorDescription">The optional extra description of the dead letter error.</param>
        /// <param name="cancellationToken">The optional <see cref="CancellationToken" /> instance to signal the request to cancel the operation.</param>
        /// <exception cref="InvalidOperationException">Thrown when the message handler was not initialized correctly.</exception>
        public async Task DeadLetterMessageAsync(string deadLetterReason, string deadLetterErrorDescription, CancellationToken cancellationToken)
        {
            await DeadLetterMessageAsync(deadLetterReason, deadLetterErrorDescription, newMessageProperties: null, cancellationToken);
        }

        /// <summary>
        /// Dead letters the Azure Service Bus message on Azure while providing <paramref name="newMessageProperties"/> for properties that has to be modified in the process.
        /// </summary>
        /// <param name="deadLetterReason">The reason why the message should be dead lettered.</param>
        /// <param name="deadLetterErrorDescription">The optional extra description of the dead letter error.</param>
        /// <param name="cancellationToken">The optional <see cref="CancellationToken" /> instance to signal the request to cancel the operation.</param>
        /// <param name="newMessageProperties">The properties to modify on the message during the dead lettering of the message.</param>
        /// <exception cref="InvalidOperationException">Thrown when the message handler was not initialized yet.</exception>
        public async Task DeadLetterMessageAsync(string deadLetterReason, string deadLetterErrorDescription, IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken)
        {
            await _messageSettle.DeadLetterMessageAsync(deadLetterReason, deadLetterErrorDescription, newMessageProperties, cancellationToken);
        }

        /// <summary>
        /// <para>
        ///     Abandon the Azure Service Bus message on Azure while providing <paramref name="newMessageProperties"/> for properties that has to be modified in the process.
        /// </para>
        /// <para>
        ///     This will make the message available again for immediate processing as the lock of the message will be released.
        /// </para>
        /// </summary>
        /// <param name="newMessageProperties">The properties to modify on the message during the abandoning of the message.</param>
        /// <param name="cancellationToken">The optional <see cref="CancellationToken" /> instance to signal the request to cancel the operation.</param>
        /// <exception cref="InvalidOperationException">Thrown when the message context was not initialized correctly.</exception>
        public async Task AbandonMessageAsync(IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken)
        {
            await _messageSettle.AbandonMessageAsync(newMessageProperties, cancellationToken);
        }

        private interface IMessageSettleStrategy
        {
            Task CompleteMessageAsync(CancellationToken cancellationToken);
            Task DeadLetterMessageAsync(string deadLetterReason, string deadLetterErrorDescription, IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken);
            Task AbandonMessageAsync(IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken);
        }

        private sealed class MessageSettleViaReceiver(ServiceBusReceiver receiver, ServiceBusReceivedMessage message) : IMessageSettleStrategy
        {
            public async Task CompleteMessageAsync(CancellationToken cancellationToken)
            {
                await receiver.CompleteMessageAsync(message, cancellationToken);
            }

            public async Task DeadLetterMessageAsync(string deadLetterReason, string deadLetterErrorDescription, IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken)
            {
                await receiver.DeadLetterMessageAsync(message, newMessageProperties, deadLetterReason, deadLetterErrorDescription, cancellationToken);
            }

            public async Task AbandonMessageAsync(IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken)
            {
                await receiver.AbandonMessageAsync(message, newMessageProperties, cancellationToken);
            }
        }

        private sealed class MessageSettleViaEventArgs(ProcessSessionMessageEventArgs messageArgs) : IMessageSettleStrategy
        {
            public async Task CompleteMessageAsync(CancellationToken cancellationToken)
            {
                await messageArgs.CompleteMessageAsync(messageArgs.Message, cancellationToken);
            }
            public async Task DeadLetterMessageAsync(string deadLetterReason, string deadLetterErrorDescription, IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken)
            {
                await messageArgs.DeadLetterMessageAsync(messageArgs.Message, newMessageProperties.ToDictionary(), deadLetterReason, deadLetterErrorDescription, cancellationToken);
            }
            public async Task AbandonMessageAsync(IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken)
            {
                await messageArgs.AbandonMessageAsync(messageArgs.Message, newMessageProperties, cancellationToken);
            }
        }
    }

    /// <summary>
    /// Represents a collection used to store properties which are set by the Service Bus service.
    /// </summary>
    public class AzureServiceBusSystemProperties
    {
        private AzureServiceBusSystemProperties(ServiceBusReceivedMessage message)
        {
            DeadLetterSource = message.DeadLetterSource;
            DeliveryCount = message.DeliveryCount;
            EnqueuedSequenceNumber = message.EnqueuedSequenceNumber;
            EnqueuedTime = message.EnqueuedTime;
            LockToken = message.LockToken;
            IsLockTokenSet = message.LockToken != null && message.LockToken != Guid.Empty.ToString();
            LockedUntil = message.LockedUntil;
            IsReceived = message.SequenceNumber > -1;
            SequenceNumber = message.SequenceNumber;
            ContentType = message.ContentType;
        }

        /// <summary>
        /// Creates an <see cref="AzureServiceBusSystemProperties"/> instance based on the available information provided in the Azure Service Bus <paramref name="message"/>.
        /// </summary>
        /// <param name="message">The received Azure Service Bus message containing the system-related information.</param>
        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="message"/> is <c>null</c>.</exception>
        public static AzureServiceBusSystemProperties CreateFrom(ServiceBusReceivedMessage message)
        {
            ArgumentNullException.ThrowIfNull(message);
            return new AzureServiceBusSystemProperties(message);
        }

        /// <summary>
        /// Gets the name of the queue or subscription that this message was enqueued on, before it was dead-lettered.
        /// </summary>
        /// <remarks>
        /// 	Only set in messages that have been dead-lettered and subsequently auto-forwarded from the dead-letter queue
        ///     to another entity. Indicates the entity in which the message was dead-lettered. This property is read-only.
        /// </remarks>
        public string DeadLetterSource { get; }

        /// <summary>
        /// Get the current delivery count.
        /// </summary>
        /// <value>This value starts at 1.</value>
        /// <remarks>
        ///    Number of deliveries that have been attempted for this message. The count is incremented when a message lock expires,
        ///    or the message is explicitly abandoned by the receiver. This property is read-only.
        /// </remarks>
        public int DeliveryCount { get; }

        /// <summary>
        /// Gets the unique number assigned to a message by Service Bus.
        /// </summary>
        /// <remarks>
        ///     The sequence number is a unique 64-bit integer assigned to a message as it is accepted
        ///     and stored by the broker and functions as its true identifier. For partitioned entities,
        ///     the topmost 16 bits reflect the partition identifier. Sequence numbers monotonically increase.
        ///     They roll over to 0 when the 48-64 bit range is exhausted. This property is read-only.
        /// </remarks>
        public long SequenceNumber { get; }

        /// <summary>
        /// Specifies if the message has been obtained from the broker.
        /// </summary>
        public bool IsReceived { get; }

        /// <summary>
        /// Gets or sets the original sequence number of the message.
        /// </summary>
        /// <value>The enqueued sequence number of the message.</value>
        /// <remarks>
        /// For messages that have been auto-forwarded, this property reflects the sequence number
        /// that had first been assigned to the message at its original point of submission. This property is read-only.
        /// </remarks>
        public long EnqueuedSequenceNumber { get; }

        /// <summary>
        /// Gets or sets the date and time of the sent time in UTC.
        /// </summary>
        /// <value>The enqueue time in UTC. </value>
        /// <remarks>
        ///    The UTC instant at which the message has been accepted and stored in the entity.
        ///    This value can be used as an authoritative and neutral arrival time indicator when
        ///    the receiver does not want to trust the sender's clock. This property is read-only.
        /// </remarks>
        public DateTimeOffset EnqueuedTime { get; }

        /// <summary>
        /// Gets the lock token for the current message.
        /// </summary>
        /// <remarks>
        ///   The lock token is a reference to the lock that is being held by the broker in ReceiveMode.PeekLock mode.
        ///   Locks are used to explicitly settle messages as explained in the <a href="https://docs.microsoft.com/azure/service-bus-messaging/message-transfers-locks-settlement">product documentation in more detail</a>.
        ///   The token can also be used to pin the lock permanently through the <a href="https://docs.microsoft.com/azure/service-bus-messaging/message-deferral">Deferral API</a> and, with that, take the message out of the
        ///   regular delivery state flow. This property is read-only.
        /// </remarks>
        public string LockToken { get; }

        /// <summary>
        /// Specifies whether or not there is a lock token set on the current message.
        /// </summary>
        /// <remarks>A lock token will only be specified if the message was received using PeekLock.</remarks>
        public bool IsLockTokenSet { get; }

        /// <summary>
        /// Gets the date and time in UTC until which the message will be locked in the queue/subscription.
        /// </summary>
        /// <value>The date and time until which the message will be locked in the queue/subscription.</value>
        /// <remarks>
        /// 	For messages retrieved under a lock (peek-lock receive mode, not pre-settled) this property reflects the UTC
        ///     instant until which the message is held locked in the queue/subscription. When the lock expires, the <see cref="P:Azure.Messaging.ServiceBus.ServiceBusReceivedMessage.DeliveryCount" />
        ///     is incremented and the message is again available for retrieval. This property is read-only.
        /// </remarks>
        public DateTimeOffset LockedUntil { get; }

        /// <summary>
        /// Gets the content type descriptor.
        /// </summary>
        /// <value>RFC2045 Content-Type descriptor.</value>
        /// <remarks>
        /// Optionally describes the payload of the message,
        /// with a descriptor following the format of RFC2045, Section 5, for example "application/json".
        /// </remarks>
        public string ContentType { get; }
    }
}
```
</details>

fgheysels: I see, it's actually a matter of just 'capturing' what varies, and abstracting those functionalities.  The implementation details are then implemented in a specific class which is just responsible for those 'implementation details'.
That's actually a very good suggestion, thanks.  I'll have a closer look at it later this week.

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
```diff
@@ -40,46 +40,38 @@ public AzureServiceBusMessageRouter(IServiceProvider serviceProvider, AzureServi
...
         /// <param name="messageContext">The context in which the <paramref name="message"/> should be processed.</param>
         /// <param name="correlationInfo">The information concerning correlation of telemetry and processes by using a variety of unique identifiers.</param>
         /// <param name="cancellationToken">The token to cancel the message processing.</param>
         /// <exception cref="ArgumentNullException">
-        ///     Thrown when the <paramref name="messageReceiver"/>, <paramref name="message"/>, <paramref name="messageContext"/>, or <paramref name="correlationInfo"/> is <c>null</c>.
+        ///     Thrown when the <paramref name="message"/>, <paramref name="messageContext"/>, or <paramref name="correlationInfo"/> is <c>null</c>.
         /// </exception>
         /// <exception cref="InvalidOperationException">Thrown when no message handlers or none matching message handlers are found to process the message.</exception>
         public async Task<MessageProcessingResult> RouteMessageAsync(
-            ServiceBusReceiver messageReceiver,
```
stijnmoreels, line 55: **requires change:** removing this on a public member is actually a breaking change that wasn't documented. To be correct, we should provide an overload without a receiver and mark the one with the receiver as deprecated.

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
```diff
@@ -210,14 +200,14 @@ Encoding DetermineEncoding()
...
-        private async Task PotentiallyAutoCompleteMessageAsync(ServiceBusReceiver messageReceiver, ServiceBusReceivedMessage message, AzureServiceBusMessageContext messageContext)
+        private async Task PotentiallyAutoCompleteMessageAsync(ServiceBusReceivedMessage message, AzureServiceBusMessageContext messageContext)
         {
             if (ServiceBusOptions.AutoComplete)
             {
                 try
                 {
                     Logger.LogTrace("Auto-complete message '{MessageId}' (if needed) after processing in Azure Service Bus in message pump '{JobId}'", message.MessageId, messageContext.JobId);
-                    await messageReceiver.CompleteMessageAsync(message);
+                    await messageContext.CompleteMessageAsync(CancellationToken.None);
```
stijnmoreels, line 210: **praise:** think indeed that it is a good way of removing the message receiver dependency, by calling the message context. Great!

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/IAzureServiceBusMessageRouter.cs
```diff
@@ -14,19 +14,15 @@ public interface IAzureServiceBusMessageRouter
...
         /// <param name="message">The incoming message that needs to be routed through registered message handlers.</param>
         /// <param name="messageContext">The context in which the <paramref name="message"/> should be processed.</param>
         /// <param name="correlationInfo">The information concerning correlation of telemetry and processes by using a variety of unique identifiers.</param>
         /// <param name="cancellationToken">The token to cancel the message processing.</param>
         /// <exception cref="ArgumentNullException">
-        ///     Thrown when the <paramref name="messageReceiver"/>, <paramref name="message"/>, <paramref name="messageContext"/>, or <paramref name="correlationInfo"/> is <c>null</c>.
+        ///     Thrown when the <paramref name="message"/>, <paramref name="messageContext"/>, or <paramref name="correlationInfo"/> is <c>null</c>.
         /// </exception>
         /// <exception cref="InvalidOperationException">Thrown when no message handlers or none matching message handlers are found to process the message.</exception>
         Task<MessageProcessingResult> RouteMessageAsync(
```
stijnmoreels, line 25: **question:** since #619 injected the concreate class of the message router, and we are actually in no need of interacting with the message router directly anymore (since we're stop supporting Azure Functions), I believe that this entire interface could be made deprecated. Though, this was not marked deprecated, so we can't technically just change this. What we can do, is deprecate the interface and not use it anymore.

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
stijnmoreels: If the suggestion in the other comment is applied, this change is not needed anymore, I believe.

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusSessionMessagePump.cs
```diff
@@ -0,0 +1,195 @@
...
+using Microsoft.ApplicationInsights;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Arcus.Messaging.Pumps.ServiceBus
+{
+    /// <summary>
+    ///     Message pump for processing messages on an Azure Service Bus entity while making use of Azure ServiceBus sessions.
+    /// </summary>
+    public class AzureServiceBusSessionMessagePump : MessagePump
```
stijnmoreels, line 19: **requires change:** a same approach as with the message context suggestion can be done here. We should also probably move down the circuit-breaker functionality from the 'main' `MessagePump` to a specific concreate class (as it shows that not all message pumps can be circuit-breaked), use the current `AzureServiceBusMessagePump` as an `abstract` class, and so forth. Similar as with the message context, also on an `internal` level.

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusSessionMessagePumpSettings.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System;
+using System.Threading.Tasks;
+using Azure.Messaging.ServiceBus;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Arcus.Messaging.Pumps.ServiceBus.Configuration
+{
+    internal class AzureServiceBusSessionMessagePumpSettings
```
stijnmoreels, line 10: **requires change:** as with PR #619 , the 'settings' class is not needed anymore, actually.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -184,9 +184,50 @@ AzureServiceBusMessagePumpSettings CreateSettings(IServiceProvider serviceProvid
...
+        /// Adds a message pump to consume messages with session-support from an Azure Service bus topic subscription.
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="topicName"></param>
+        /// <param name="subscriptionName"></param>
+        /// <param name="fullyQualifiedNamespace"></param>
+        /// <param name="credential"></param>
+        /// <param name="configureMessagePump"></param>
+        /// <returns></returns>
+        public static ServiceBusMessageHandlerCollection AddServiceBusTopicMessagePumpWithSessionSupport(
```
stijnmoreels, line 197: **question:** the whole exercise previously was to reduce the number of ways to register message pumps and solely focus on a limit number of them. Adding them now for internal details decisions, seems like making the same mistake as we did previously.
The `configureMessagePump` is already another parameter type, which means that this could be an overload, in fact...

src/Arcus.Messaging.Abstractions/MessageHandling/MessageHandler.cs
fgheysels: But why would you keep this additional check, if it's already covered by the check on L131 as well ?  (Ok, the check that is removed checks if the given context is the exact type, whereas the check on L131 will also evaluate to true when using a subclassed MessageContext.  (And I don't think that's an issue tbh)

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
stijnmoreels: Yes, of course, not really what I was trying to say. It is still something that should be clear from registering the message pump, just that another extension set - that blows up again the amount of extensions/overloads necessary, might not be the way forward, as it is a 'Service Bus' decision, not a 'message pump' decision, per se ➡️ as shown in the highly duplicated code.

It could be something configured in the options, for example:
```csharp
services.AddServiceBusTopicMessagePump(..., options =>
{
    // ...returns the session-specific options
    options.UseSesssions(session => session.SessionIdleTimeout = TimeSpan.FromSeconds(30));
});
```

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/IAzureServiceBusMessageRouter.cs
fgheysels: Is this something we should do in the scope of this PR ?

src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusSessionMessagePump.cs
fgheysels: And indeed, the CircuitBreaker functionality is currently also something that is hindering us from doing the suggested change.
I believe that doing this refactoring would be a bit too much scope-creep for this PR tbh.

src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusSessionMessagePumpSettings.cs
fgheysels: Indeed, I've removed this class.

src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
```diff
@@ -135,8 +157,156 @@ public async Task DeadLetterMessageAsync(string deadLetterReason, string deadLet
...
+
+            public Task DeadLetterMessageAsync(string deadLetterReason, string deadLetterErrorDescription, IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken)
+            {
+                if (newMessageProperties != null)
+                {
+                    var propertiesToModify = newMessageProperties.ToDictionary();
+                    return _args.DeadLetterMessageAsync(_args.Message, propertiesToModify, deadLetterReason, deadLetterErrorDescription, cancellationToken);
+                }
+
+                return _args.DeadLetterMessageAsync(_args.Message, deadLetterReason, deadLetterErrorDescription, cancellationToken);
```
fgheysels, line 300: This is a strange construction, but this is due to some weird implementation choice in the ServiceBus SDK I believe.
The `ProcessSessionMessageEventArgs.DeadLetterMessageAsync` has [these](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.processsessionmessageeventargs.deadlettermessageasync?view=azure-dotnet) overloads.
Note the fact that one method accepts a `Dictionary` and another one accepts an `IDictionary` (that allows nulls, as the default value for that parameter is null).
When the `DeadLetterMessageAsync` method that accepts a `Dictionary` is called, and you pass `null` to that parameter, a `NullReferenceException` is being thrown.

src/Arcus.Messaging.Abstractions.ServiceBus/AzureServiceBusMessageContext.cs
fgheysels: This is a much better solution indeed, thanks!

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/AzureServiceBusMessageRouter.cs
fgheysels: True, but is it worth the fuzz ?
What's the chance that someone has created his own message-pump and uses the MessageRouter, or uses the MessageRouter directly ?  In all honesty, I think chances are very slim.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
fgheysels: As discussed offline, you can take over this PR to finalize the PR, as I'm going on holiday and the amount of time I can spend on this PR will be limited in the first few weeks after my holiday.

PR #600, opened 2025-06-24
chore(az.sb): remove deprecated az service-bus message pump options

stijnmoreels:
> 👉 Since we're working on v3.0, we can start removing stuff for real.

This PR removes the deprecated customization functionality via dedicated options or (already) removed overloads (configuration connection string, Arcus secret store, automatic topic creation...).

Relates to #484


src/Arcus.Messaging.Pumps.ServiceBus/Configuration/AzureServiceBusMessagePumpOptions.cs
```diff
@@ -152,14 +93,13 @@ public int MaximumUnauthorizedExceptionsBeforeRestart
...
+#pragma warning restore CS0618 // Type or member is obsolete
 
         /// <summary>
         /// Gets the default consumer-configurable options for Azure Service Bus Queue message pumps.
         /// </summary>
-        internal static AzureServiceBusMessagePumpOptions DefaultOptions => new AzureServiceBusMessagePumpOptions()
-        {
-            TopicSubscription = ServiceBus.TopicSubscription.None
-        };
+        internal static AzureServiceBusMessagePumpOptions DefaultOptions => new();
```
fgheysels, line 103: It's a detail, but I'm just wondering if an `internal static readonly` member isn't better here then an `internal static` property ?
The semantics are somewhat different I guess, where an `internal static readonly` is only initialized once, and is only initialized when the member is accessed for the first time.
With a property implementation like it is now, I think we'll be initializing and returning a new instance every time we access the property.

fgheysels: Why, I would assume that this is just a (readonly) instance that just contains the default settings ?  Why would `DefaultOptions` be mutable ?

stijnmoreels: Because it is passed as a 'default/initial' value for the `Action<...> configureOptions` parameter. The instance can get mutated there by the user.

fgheysels (approved):

PR #610, opened 2025-06-26
chore(auto.complete): move az sb auto-complete func to message router

### LINKED ISSUE #596: 🕵️ [Investigate]: whether we can complete the message from the Azure Service Bus message router
Currently, the received message is completed - upon user's requested options - by the Azure Service Bus message pump. Since other message settle operations are done in the router (dead-letter, abandon), we should investigate if we can also complete the message in the message router.

First noted #593

stijnmoreels:
It came to our attention that no all message settlement operations were located at the Azure Service Bus message router (i.e. message auto-completion). After a closer inspection, this functionality can be easily moved to the router, making both the message pump and message router more clear in their responsibilities.

Closes #596


fgheysels (approved):

stijnmoreels:
> It's still in draft, but I already approved the PR. What changes are still to be done ?

Possibly could have some merge conflicts with other PR's that were merged, but everything now seems fine.

PR #630, opened 2025-07-19
feat: remove contenttype property from PropertyNames

joachimgoris:


src/Arcus.Messaging.Tests.Integration/MessagePump/ServiceBusMessagePumpTests.cs
```diff
@@ -49,7 +49,57 @@ public async Task ServiceBusMessagePumpWithQueueAndTopic_PublishServiceBusMessag
...
+                                   .Cast<AzureServiceBusMessagePump>()
+                                   .Select(m => m.EntityType)
+                                   .ToArray());
+
+            return registeredEntityType;
+        }
+
+        private static ServiceBusMessage CreateOrderServiceBusMessageForW3C(Encoding encoding = null)
+        {
+            encoding ??= Encoding.UTF8;
```
stijnmoreels, line 65: **requires chagne:** these are some weird additions that probably happened during an incorrect merge with `main`. Best that these are removed.

stijnmoreels (changes requested):
Thx for helping out, only an incorrect merge with `main` caused both the build to fail as well as the PR changes to corrupt.

PR #635, opened 2025-07-30
feat(az.sb.session): add session support for az service-bus message pump

### LINKED ISSUE #489: Support Session-enabled topics and queues in Arcus.Messaging

stijnmoreels:
Provide with an easy `options.UseSessions(...)` method on the options the possibility to run the Azure Service Bus message pump as a session-aware pump (currently w/o circuit breaker functionality).

Closes #489 
Alternative PR on old #597


src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -56,7 +56,7 @@ public static ServiceBusMessageHandlerCollection AddServiceBusQueueMessagePump(
             string queueName,
             string fullyQualifiedNamespace,
             TokenCredential credential,
-            Action<AzureServiceBusMessagePumpOptions> configureMessagePump)
+            Action<ServiceBusMessagePumpOptions> configureMessagePump)
```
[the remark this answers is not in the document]
stijnmoreels: Yes/no, I believe, since renaming it is on the planning and this functionality requires us to change the options. (Bit like the pumps are also renamed). But happy to move to separate PR, if need be.

src/Arcus.Messaging.Pumps.ServiceBus/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -201,19 +201,25 @@ private static ServiceBusMessageHandlerCollection AddServiceBusMessagePump(
...
 
-            return new ServiceBusMessageHandlerCollection(services) { JobId = options.JobId };
+            return new ServiceBusMessageHandlerCollection(services)
+            {
+                JobId = options.JobId,
+                UseSessions = options.RequestedToUseSessions
+            };
         }
 
         private static string SanitizeSubscriptionName(string subscriptionName, IServiceProvider provider)
```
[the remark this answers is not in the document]
stijnmoreels: Ah, yes, think indeed that we can remove this one.
Added an issue for it here: https://github.com/arcus-azure/arcus.messaging/issues/640

src/Arcus.Messaging.Abstractions.ServiceBus/MessageHandling/ServiceBusMessageHandlerCollection.cs
```diff
@@ -18,5 +18,10 @@ public class ServiceBusMessageHandlerCollection : MessageHandlerCollection
         public ServiceBusMessageHandlerCollection(IServiceCollection services) : base(services)
         {
         }
+
+        /// <summary>
+        /// Gets or sets the boolean flag to indicate whether the message pump is configured to use sessions.
+        /// </summary>
+        public bool UseSessions { get; set; }
```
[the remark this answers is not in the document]
fgheysels: Ok, we'll leave it as is.

src/Arcus.Messaging.Pumps.ServiceBus/ServiceBusMessageRouter.cs
```diff
@@ -0,0 +1,197 @@
...
+                                .FirstOrDefault(e => e.Name.Equals(encodingName, StringComparison.OrdinalIgnoreCase));
+
+                    return foundEncoding?.GetEncoding() ?? fallbackEncoding;
+                }
+
+                return fallbackEncoding;
+            }
+        }
+
+        private async Task PotentiallyAutoCompleteMessageAsync(ServiceBusReceivedMessage message, AzureServiceBusMessageContext messageContext)
```
fgheysels, line 159: I believe we can remove the `message` parameter ?  Isn't everything available through he `messageContext` parameter ?

stijnmoreels: Done here: https://github.com/arcus-azure/arcus.messaging/pull/635/commits/69827bd7dc6de6f3347fe1e1794374b91be75ece

src/Arcus.Messaging.Pumps.ServiceBus/ServiceBusMessageRouter.cs
```diff
@@ -0,0 +1,197 @@
...
+                try
+                {
+                    Logger.LogTrace("Auto-complete message '{MessageId}' (if needed) after processing in Azure Service Bus in message pump '{JobId}'", messageContext.MessageId, messageContext.JobId);
+                    await messageContext.CompleteMessageAsync(CancellationToken.None);
+                }
+                catch (ServiceBusException exception) when (
+                    exception.Message.Contains("lock")
+                    && exception.Message.Contains("expired")
+                    && exception.Message.Contains("already")
+                    && exception.Message.Contains("removed"))
```
fgheysels, line 172: I think it's a bit 'dangerous' to filter on (sub)strings that are part of the exception-message.
I believe we should have a look if we can't make use of the `Reason` property of the `ServiceBusException` class.

fgheysels (approved):

src/Arcus.Messaging.Pumps.ServiceBus/ServiceBusMessageRouter.cs
stijnmoreels: It was done in the session pump, but not in the receiver pump.

stijnmoreels:
> Great work! I am eager to test it out.

@fgheysels , building a [v3.0.0-preview-1](https://dev.azure.com/codit/Arcus/_build/results?buildId=229774&view=results) preview for you.

PR #638, opened 2025-07-31
chore(messaging.abstractions): make messagecontext protected abstract

joachimgoris:
Multiple changes to class `MessageContext`:
- Make class abstract.
- Remove obsolete constructor.
- Make remaining constructor `protected`.
- Remove `MessageContextTests` due to inaccessibility.

### DIFF
--- a/src/Arcus.Messaging.Abstractions/MessageContext.cs
+++ b/src/Arcus.Messaging.Abstractions/MessageContext.cs
@@ -21,10 +21,11 @@ namespace Arcus.Messaging.Abstractions
         {
             ArgumentException.ThrowIfNullOrWhiteSpace(messageId);
             ArgumentException.ThrowIfNullOrWhiteSpace(jobId);
+            ArgumentNullException.ThrowIfNull(properties);
 
             MessageId = messageId;
             JobId = jobId;
-            Properties = properties ?? throw new ArgumentNullException(nameof(properties));
+            Properties = properties;
         }
 
         /// <summary>


src/Arcus.Messaging.Abstractions/MessageContext.cs
```diff
@@ -35,7 +17,7 @@ public MessageContext(string messageId, IDictionary<string, object> properties)
         /// <param name="properties">The contextual properties provided on the message.</param>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="messageId"/> or the <paramref name="jobId"/> is blank.</exception>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="properties"/> is <c>null</c>.</exception>
-        public MessageContext(string messageId, string jobId, IDictionary<string, object> properties)
+        protected MessageContext(string messageId, string jobId, IDictionary<string, object> properties)
         {
             if (string.IsNullOrWhiteSpace(messageId))
```
stijnmoreels, line 22: Since we're on this class anyway, we can also use the new `ArgumentException.Throw...` extensions to validate the incoming parameters.

src/Arcus.Messaging.Tests.Unit/MessageContextTests.cs
```diff
@@ -1,90 +0,0 @@
-﻿using System;
-using System.Collections.Generic;
-using Arcus.Messaging.Abstractions;
-using Xunit;
-
-namespace Arcus.Messaging.Tests.Unit
-{
-    public class MessageContextTests
```
stijnmoreels, line 8: **info:** normally, these tests would have to be moved to the concrete type, instead of 'removing due to inaccessibility' -> which is not a good reason. Though, in this case, the tests have implicitly been covered now with the arrival of other functionality, so we can ignore them safely.

stijnmoreels (changes requested):
Thx for this!

src/Arcus.Messaging.Abstractions/MessageContext.cs
joachimgoris: All arguments of that method are now being checked by the new extension methods instead of the oldschool way. I will also need new tasks or issues to pick up :)

PR #649, opened 2025-09-09
feat(az.sb): add startup hook for az sb message pump

### LINKED ISSUE #504: ✨ [New]: possibility to explicitly start the MessagePump
When using Arcus.Messaging, we typically add a MessagePump to our DI container by calling one of the extension methods that adds a message-pump.  This method adds a `BackGroundService` to our Dependency Container.

As far as I know or can see, the BackGroundJob immediately starts, but it would be good if an application programmer could have a possibility to decide when the MessagePump should in fact start reading from ServiceBus.

I have a scenario where I want to make sure that certain dependencies are available before I want the MessagePump to start receiving messages.   AFAIK, I can't use the service-breaker functionality for this, because the service-breaker can only be activated once we've already received a message.

I don't know if it would be possible to register / inject a class that implements a certain interface that an application programmer could use to implement some logic.  Only when that logic is executed, the MessagePump starts reading from the broker.
That class could then maybe be called by the `StartProcessingMessagesAsync` method of the `MessagePump` class

Image

stijnmoreels:
Adds a function registration to the Azure Service Bus message pump that adds a 'hook' before starting up and receiving messages. Useful for when dependent systems are not always directly available.

Closes #504


src/Arcus.Messaging.Pumps.ServiceBus/Configuration/ServiceBusMessagePumpOptions.cs
```diff
@@ -152,4 +158,27 @@ public TimeSpan SessionIdleTimeout
...
+    public class ServiceBusMessagePumpHooksOptions
+    {
+        internal Func<IServiceProvider, Task> BeforeStartupAsync { get; set; } = _ => Task.CompletedTask;
+
+        /// <summary>
+        /// Sets a function on the message pump that should run before the pump receives Azure Service Bus messages.
+        /// Useful for when dependent systems are not always directly available.
+        /// </summary>
+        /// <remarks>
+        ///     ⚠️ Multiple calls will override each other.
```
fgheysels, line 174: Wouldn't it be nice if multiple calls would result in multiple hooks being executed ?
It would make it all more complex of course ....

stijnmoreels: Yes, would they need to be run serial, parallel, should only one be finished or all...

fgheysels: Indeed, let's keep it as is.

PR #650, opened 2025-09-09
docs(az.sb): Update Autocomplete documentation on Service Bus Arcus.Messaging

gverstraete:
Adds warning on `options.AutoComplete` to point to the fact that consumers are responsible for completing the message themselves, if not auto-completing it.


stijnmoreels (approved):
A bit of clarification never hurts. Thx!

PR #669, opened 2025-10-16
test(routing): add new in-depth unit tests for general message routing

stijnmoreels:
Adds unit tests (both example-based and property-based) that tests in-depth the general message routing functionality. Because of these tests, some edge-case issues were found and fixed here as well. Mostly related to matching functionality and correct error summary logging.

> [!NOTE]  
> The example-based unit tests are for human consumption and are mostly there to showcase in the most simple manner how the message routing functionality works. The property-based unit tests are there to in-depth generate 100 tests / property so that we can randomize scenarios and be more certain of stable functionality. Combining them is key in test communication.

These tests will act as a substitute for the current ones that verify only deprecated functionalty.

### DIFF
--- a/src/Arcus.Messaging.Core/MessageHandling/MessageHandler.cs
+++ b/src/Arcus.Messaging.Core/MessageHandling/MessageHandler.cs
@@ -3,6 +3,8 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Reflection;
+using System.Text;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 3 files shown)


src/Arcus.Messaging.Core/MessageHandling/MessageHandler.cs
```diff
@@ -237,15 +237,8 @@ private static ProcessMessageAsync DetermineMessageImplementation<TMessage, TMes
...
-                if (rawMessage is not TMessage message)
-                {
-                    return MessageProcessingResult.Failure(generalMessageContext.MessageId, MessageProcessingError.MatchedHandlerFailed, $"requires message type {typeof(TMessage).Name}");
-                }
-
-                if (generalMessageContext is not TMessageContext messageContext)
-                {
-                    return MessageProcessingResult.Failure(generalMessageContext.MessageId, MessageProcessingError.MatchedHandlerFailed, $"requires message context type {typeof(TMessageContext).Name}");
-                }
+                var message = (TMessage) rawMessage;
```
fgheysels: Isn't it possible that this results in a cast exception ?
(Previous code checked if the received message was of type `TMessage` ?

stijnmoreels: It was moved to another place. The `MessageRouter` always first checks if it matches before going through this one, if it would be added here as well it would be duplication, but I agree, it is not the best design-wise.
Let's see if we can make this more rational..

PR #673, opened 2025-11-04
chore(az.sb): rename 'max concurrent calls' to 'max messages / batch' option

### LINKED ISSUE #659: Rename MaxConcurrentCalls property to MaxMessages
Our `ServiceBusMessageOptions` class contains a parameter called `MaxConcurrentCalls`.  This option defines the amount of messages that will be retrieved in one batch from Azure ServiceBus.  

This property name has been choosen based on the `MaxConcurrentCalls` property that exists in the ServiceBus binding of Azure Functions, however, this name is confusing.  It would be better to rename this property to `MaxMessages` which better reveals the purpose of this property.  (Moreover, in our pump implementation, this property is passed as the `maxMessages` argument in the call to the `ReceiveMessagesAsync` method of the `ServiceBusReceiver`).

We can introduce a new `MaxMessages` property in the `ServiceBusMessageOptions` and deprecate the exising `MaxConcurrentCalls` property.  The existing `MaxConcurrentCalls` property could also delegate to the new `MaxMessages` property.

stijnmoreels:
Rename delegate the `MaxConcurrentCalls` Azure Service Bus message pump option to a new `MaxMessagesPerBatch` option and deprecate the former.

Closes #659

### DIFF
--- a/src/Arcus.Messaging.ServiceBus/Configuration/ServiceBusMessagePumpOptions.cs
+++ b/src/Arcus.Messaging.ServiceBus/Configuration/ServiceBusMessagePumpOptions.cs
@@ -29,7 +29,7 @@ namespace Arcus.Messaging.Pumps.ServiceBus.Configuration
         /// <summary>
         /// Gets the amount of messages that will be retrieved in one batch from Azure Service Bus.
         /// </summary>
-        /// <remarks>The default value is 1</remarks>
+        /// <remarks>The default value is 1.</remarks>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="value"/> is less than or equal to zero.</exception>
         public int MaxMessagesPerBatch
         {


src/Arcus.Messaging.ServiceBus/ServiceBusSessionMessagePump.cs
```diff
@@ -58,6 +58,7 @@ private ServiceBusSessionProcessor CreateSessionProcessor()
             {
                 PrefetchCount = Options.PrefetchCount,
                 MaxConcurrentSessions = Options.Session.MaxConcurrentSessions,
+                MaxConcurrentCallsPerSession = Options.Session.MaxConcurrentCallsPerSession,
```
stijnmoreels: Eagle eye strikes again.

src/Arcus.Messaging.ServiceBus/Configuration/ServiceBusMessagePumpOptions.cs
```diff
@@ -19,13 +19,25 @@ public class ServiceBusMessagePumpOptions
...
+            get => MaxMessages;
+            set => MaxMessages = value;
+        }
+
+        /// <summary>
+        /// Gets the amount of messages that will be retrieved in one batch from Azure Service Bus.
+        /// </summary>
+        /// <remarks>The default value is 1</remarks>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="value"/> is less than or equal to zero.</exception>
+        public int MaxMessages
```
fgheysels: I know I've coined `MaxMessages` but when I read the comments, maybe we should change it to `MaxMessagesPerBatch` or `MessagesPerBatch` ? wdyt ?

src/Arcus.Messaging.ServiceBus/ServiceBusSessionMessagePump.cs
fgheysels: Ow, that's an important one!

src/Arcus.Messaging.ServiceBus/Configuration/ServiceBusMessagePumpOptions.cs
fgheysels: thanks!

PR #703, opened 2026-04-27
Remove upper-version dependency constraints

### LINKED ISSUE #702: Remove upper boundary version-constraints of dependent packages
Arcus.Messaging uses lower and upper-bound version constraints on certain packages.
Specifying upper bound version constraints bites us in the back when using Arcus in real projects.   It's better to just remove those upper bound version constraints.

fgheysels:
Remove the specified max-version dependency constraints.
Updated the Moq package to the latest version in the unit-test project as well, as the specified version contained vulnerabilities.

Closes #702

### DIFF
--- a/src/Arcus.Messaging.Core/Arcus.Messaging.Core.csproj
+++ b/src/Arcus.Messaging.Core/Arcus.Messaging.Core.csproj
@@ -32,8 +32,8 @@
   </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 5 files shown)


stijnmoreels (commented):
We should not do this, as this makes supporting and maintaining of the library even harder, as we've seen in the past.

fgheysels:
I see no issues in removing dependency version constraints that are just there to 'possibly avoid issues that might occur'.  There are no proven issues right now with version higher than the defined constraints.

PR #704, opened 2026-04-30
chore(deps): use correct dependency constraints for .NET 10 target framework

stijnmoreels:
* Use correct 11 upper threshold boundary for the `System.Data.Memory` package as this will include all the .NET 10-related minor versions.
* Updated Moq library to most recent version.

### DIFF
--- a/src/Arcus.Messaging.Core/Arcus.Messaging.Core.csproj
+++ b/src/Arcus.Messaging.Core/Arcus.Messaging.Core.csproj
@@ -33,7 +33,7 @@
 
     <ItemGroup>
         <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="[8.*,11.0.0)" />
-        <PackageReference Include="System.Memory.Data" Version="[8.*,10.0.0)" />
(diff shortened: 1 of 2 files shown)


fgheysels:
I believe there's also an issue with the constraints on Azure.Messaging.ServiceBus in Arcus.Messaging.ServiceBus.csproj.

And there's also a vulnerability in the version that is used for the Moq package in the unit-test project.

stijnmoreels:
> I believe there's also an issue with the constraints on Azure.Messaging.ServiceBus in Arcus.Messaging.ServiceBus.csproj.

The most recent version of that package is v7.20.1, and we're within that boundary. When we release a new version, it will take probably a more recent version that will resolve the issue you have.

> And there's also a vulnerability in the version that is used for the Moq package in the unit-test project.

Will not affect the release, I believe, since we do not fail on warnings based on test projects; but updated to most recent version here as well.

fgheysels (approved):

fgheysels:
Thanks for the necessary changes.
I believe we can now build a new release ?

PR #708, opened 2026-06-18
feat(az.sb): Idempotent message-settlement

### LINKED ISSUE #707: 💪 [Enhance]: avoid possible MessageLockLostExceptions being thrown during message-settlement
In a project where we're using Arcus.Messaging, we see a lot of 'ServiceBusExceptions' are being thrown.
This happens in situations where we deadletter a ServiceBusReceivedMessage in our MessageHandler and AutoComplete.
I think Arcus can be improved a little bit by keeping track of the 'settlement-state' in the IMessageSettleStrategy implementations.
The IMessageSettleStrategy can keep track of a boolean flag (`_messageAlreadySettled` or something like that).  When `CompleteMessageAsync` , `DeadletterMessageAsync` or `AbandonMessageAsync` is called, that flag can be set.
Those methods should also check if the flag hasn't been set before doing the actual settlement.

I think this can avoid a lot of exceptions being thrown and logged to application-insights which will have a positive effect on resource usage and Azure consumption cost.

fgheysels:
Avoid that ServiceBusExceptions are thrown when message-settlement is executed multiple times on the same ServiceBusMessageContext.

Closes #707

### DIFF
--- a/src/Arcus.Messaging.ServiceBus/AzureServiceBusMessageContext.cs
+++ b/src/Arcus.Messaging.ServiceBus/AzureServiceBusMessageContext.cs
@@ -20,7 +20,7 @@ namespace Arcus.Messaging.ServiceBus
             string fullyQualifiedNamespace,
             ServiceBusEntityType entityType,
             string entityPath,
-            IMessageSettleStrategy messageSettle,
+            MessageSettleStrategy messageSettle,
             ServiceBusReceivedMessage message)
             : base(message.MessageId, jobId, message.ApplicationProperties.ToDictionary(item => item.Key, item => item.Value))
         {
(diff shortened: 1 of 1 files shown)


src/Arcus.Messaging.ServiceBus/AzureServiceBusMessageContext.cs
```diff
@@ -109,17 +109,70 @@ public static ServiceBusMessageContext Create(
...
+
+            public Task AbandonMessageAsync(IDictionary<string, object> newMessageProperties, CancellationToken cancellationToken)
+            {
+                if (_messageIsSettled)
+                {
+                    return Task.CompletedTask;
+                }
+
+                var result = AbandonMessageCoreAsync(newMessageProperties, cancellationToken);
+                _messageIsSettled = true;
```
stijnmoreels, line 165: **note:** since this is set before `await`, it could still fail during settlement, and incorrectly set as 'settled'.

fgheysels: Ah, indeed.  Thanks for pointing that out.
So, what do you suggest ?  Using await here ?

stijnmoreels: Yes, think so.

fgheysels: I made the change.  Stupid of me that I didnt notice this. Thanks for the awareness!

stijnmoreels (approved):

